PR #6, opened 2018-01-19
Completed up to Connected Components

47analogy:
will try to implement some of the stretch goals


beejjorgensen:
I see the BFS in there and the drawing code looks excellent, but I don't see the connected components code... am I missing it?

Well done on the stuff that's there, though. :)

PR #35, opened 2018-03-29
Jesse Hood - all phases + stretches 1 and 2

jessehood:


jessehood:
Forgot to randomize the weights for stretch 1 but got them to render at least.

PR #70, opened 2018-05-08
Christopher Beards

ChristopherBeards:


wildpow:
Like always you have regular commits, with clean and clear messages each step of the way Also you're right on track with your implementation. You are a gentleman and a scholar sir.

PR #139, opened 2018-07-09
Graphs Justin H

JustinSHong:
Implemented Edge, Vertex, bfs in graph.js
added vertex/edge drawing functionality in updateCanvas()


JustinSHong:
implemented bfs() and getConnectedComponents() to find components and assign each component a random color

implemented drawing logic for edges

PR #162, opened 2018-07-23
Il Ho Kim

ilhoKim:
- add non-project folders to root .gitignore


nikhilkamineni:
Nice job getting the labels working.

I would probably add some code in your `draw.py` file so that when you run that file (e.g. `python draw.py` in your terminal) it automatically generate a graph. This will make development go quicker than having to open up the repl and manually do it.

PR #172, opened 2018-07-23
Andrew Kolumbic

AKolumbic:


AKolumbic:
adds draw.py commit includes everything from today's lecture, it's in working order, when I run it directly from the command line, but I don't believe the main() functions are working correctly, or if they are correct, I'm executing it wrong.

AKolumbic:
I added comments and print() statements throughout my BFS alg to explain what I was doing and to have proof that it works as intended(up to the point where it doesn't)

PR #176, opened 2018-07-23
Ja Navey

JSNavey:


nikhilkamineni:
Nice work so far. Ran your code on my machine and it generated a graph correctly with labels.

Your BFS algorithm looks correct. The next step would be to try and figure how to use it to find connected components (you'll probably want to increase the default number of nodes/vertices your graph randomly generates to be able to test this).

nikhilkamineni:
https://github.com/JSNavey/Graphs/blob/1a3d8b702dc3f017755582497deddf49a26ec5e7/projects/graph/src/graph.py#L91

It would be better (i.e more 'pythonic') to write this line as `if not visited[int(i)]:`

PR #184, opened 2018-07-23
Anne Courtney

annecourtney:


projects/graph/src/graph.py
```diff
@@ -1,9 +1,157 @@
...
+        self.vertices = {}
+        self.components = 0
+
+    def add_vertex(self, vertex):
+        if not isinstance(vertex, Vertex):
+            vertex = Vertex(vertex)
+
+        if vertex.label in self.vertices:
+            return False  # ignores it
+
```
rawalpunit: Why not raise an exception so you know that the vertex was already existing?

projects/graph/src/graph.py
```diff
@@ -1,9 +1,157 @@
...
+        if start not in list(self.vertices.keys()):
+            self.add_vertex(Vertex(start))
+        # adding the end point if it doesn't exist in the vertices list/array already
+        if end not in list(self.vertices.keys()):
+            self.add_vertex(Vertex(end))
+
+        self.vertices[start].edges.add(self.vertices[end])
+        if multidirectional:
+            self.vertices[end].edges.add(self.vertices[start])
+
```
rawalpunit: What happens if the start and/or end already exist?

projects/graph/src/graph.py
```diff
@@ -1,9 +1,157 @@
...
+    #     while stack:
+    #         current = stack.pop()
+    #         if current == target:
+    #             break
+    #         visited.add(current)
+    #         # try to add possible unvisited vertices to the queue
+    #         stack.extend(self.vertices[current] - visited)
+
+    #     return visited
+
```
rawalpunit: nice job here!

projects/graph/src/graph.py
```diff
@@ -1,9 +1,157 @@
...
+    #     while queue:
+    #         current = queue.pop(0)
+    #         if current == target:
+    #             break
+    #         visited.add(current)
+    #         # try to add possible unvisited vertices to the queue
+    #         queue.extend(self.vertices[current] - visited)
+
+    #     return visited
+
```
rawalpunit: nice clean code

PR #187, opened 2018-07-23
Maribel Cuales

maribelcuales:
## Drawing Connected Components


nikhilkamineni:
Nice progress so far.

When I run your draw.py file I get two graphs being generated. Is that intentional?

nikhilkamineni:
https://github.com/maribelcuales/Graphs/blob/5f28843464022fa8baa63425ae39ed73243e1c0a/projects/graph/src/draw.py#L89

This import is unnecessary since this is in your draw.py file (you would only need that import if you were in a different file).

PR #190, opened 2018-07-23
Erinç Emer

erinc35:


nikhilkamineni:
Nice progress so far. Cool to see you figured out how to add arrows. I would probably put that logic in it's own method on your BokeGraph class though. It will be easier read when someone else (or you in the future) comes looks at this code.

PR #196, opened 2018-07-23
Brandon Benefield

bbenefield89:


nikhilkamineni:
Nice work so far. 

I would advise against naming the dict storing your vertices on your Graph class 'graph'. For someone else (or you in the future) looking at this code, seeing something like `graph.graph` is not the most intuitive, especially since your class is also called `Graph`.

I'd also advise against creating an instance of `Graph` in `graph.py` and importing both the class and that instance into `draw.py`. It's usually best to have separation of concerns so creating an instance of `Graph` would be better done inside `draw.py` (or even another file entirely).

Also, be sure to eventually put all of the stuff in `draw.py` inside the BokehGraph class.

I understand some of the stuff I mentioned is due to hacking away and trying to get things working. and that is totally fine. This is just some stuff to keep in mind when you get to the point where you want to refactor.

PR #235, opened 2018-09-10
Blake Fletcher - Graphs

blkfltchr:


blkfltchr:
Need to look over the code and annotate for better understanding, and refactor graph.py so the earlier code works and new code exists on graph_demo.py

PR #237, opened 2018-09-10
Thomas Greenhalgh - Graphs

tgreenhalgh:


projects/graph/src/graph.py
```diff
@@ -11,7 +11,10 @@ def __init__(self):
         self.vertices = {}
 
     def add_vertex(self, vertex):
-        self.vertices[vertex] = set()
+        if vertex not in self.vertices:
+            self.vertices[vertex] = set()
+        else:
+            sys.exit("Vertex {} already exists.".format(vertex))
```
Supertatortots: Good thinking on checking if the Vertex is there!

PR #263, opened 2018-10-15
UNICORN Carey Baldwin

careybaldwin02:
Just getting started - working on creating a graph class.

### DIFF
--- a/projects/graph/src/graph.py
+++ b/projects/graph/src/graph.py
@@ -7,3 +7,16 @@ class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
         pass  # TODO
+
+
+
+
+# Tests for Graph Class
+graph = Graph()  # Instantiate your graph
+graph.add_vertex('0')
+graph.add_vertex('1')
+graph.add_vertex('2')
+graph.add_vertex('3')
+graph.add_edge('0', '1')
+graph.add_edge('0', '3')
+print(graph.vertices)
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.linting.pylintEnabled": false
+}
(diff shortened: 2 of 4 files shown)


projects/graph/src/graph.py
```diff
@@ -2,8 +2,117 @@
...
+    
+    def __contains__(self,n):
+        return n in self.vertices
+
+    def add_edge(self, start, end, bothWays = True):
+        if start not in self.vertices or end not in self.vertices:
+            raise Exception('These vertices are not in the graph')
+        self.vertices[start].add(end)
+        if bothWays:
+            self.vertices[end].add(start)
```
Mister-Corn: I appreciate that you have additional arguments for `add_vertex` and `add_edge` to accommodate additional information :+1:

projects/graph/src/graph.py
```diff
@@ -2,8 +2,117 @@
...
+
+    def vertices(self):
+        return self.vertices.keys()
+
+    def BFS(self, s):
+    # https://www.geeksforgeeks.org/breadth-first-search-or-bfs-for-a-graph/
+        self.graph = defaultdict(list)
+        # mark all of the vertices as not visited
+        visited = [False] * (len(self.graph))
```
Mister-Corn: This is a very great way to keep track of visited nodes. Another common way is to make `visited` an attribute for a given Vertex object, set it false, and toggle it when needed. I feel this way might be better, but it's worth thinking about performance and practical considerations of each approach.

projects/graph/src/graph.py
```diff
@@ -2,8 +2,117 @@
...
+
+        # mark the source node as visited and enqueue
+        queue.append(s)
+        visited[s] = True
+
+        while queue:
+
+            # Dequeue a vertex from the queue and print it
+            start = queue.pop(0)
+            print (s, end = " ")
```
Mister-Corn: If you have an eye for performance optimizations, consider researching `deque`'s. They are built in data structure which are like lists but have interesting properties you may find useful.

careybaldwin02:
Thanks for the comments!  All super helpful!

PR #264, opened 2018-10-15
Keegan Burkett: Open PR

keegn:
## Day 1 
- [ ] Undirected graph of 4 verts.
- [ ] Directed graph of 5 verts.
- [ ] Cyclic directed graph of 6 verts.
- [ ] DAG of 7 verts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,3 +11,4 @@
 ## Projects
 
 * [Drawing Connected Components](projects/graph)
+* 


projects/graph/src/graph.py
```diff
@@ -6,4 +6,20 @@
...
+            self.vertices[v2].edges.add(v1)
+        else: 
+            reaise IndexError("That vertex does not exist")
+
+class Vertex:
+    def __init__(self, vertex_id):
+        self.id = vertex_id
+        self.edges = set()
+    def __repr__(self):
+        return (f'{self.edges}')
```
honda0306: I like your use of the fstring here. Now that you have the Vertex class up and running, things should go much faster and smoother from here on.

PR #265, opened 2018-10-15
Andrew McLaughlin (Graphs)

LaikaFusion:


projects/graph/src/graph.py
```diff
@@ -6,4 +6,12 @@
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
-        pass  # TODO
+        self.vertices = dict()
+    def add_vertex(self,vertex):
+        self.vertices[vertex] = set()
```
kelfro: You may want to add a conditional here to raise an Exception if the vertex already exists, or else you might end up with overlapping duplicates.

projects/graph/src/draw.py
```diff
@@ -10,5 +10,36 @@
...
+                y.append(num)
+        graph.edge_renderer.data_source.data = dict(start=x,end=y)
+
+        circ = [i*2*math.pi/8 for i in node_indices]
+        xcir = [math.cos(i) for i in circ]
+        ycir = [math.sin(i) for i in circ]
+        graph_layout = dict(zip(node_indices, zip(xcir, ycir)))
+        graph.layout_provider = StaticLayoutProvider(graph_layout=graph_layout)
+        source = ColumnDataSource(dict(xvals=xcir,yvals=ycir,name=node_indices))
+
```
kelfro: Look at all this math here! Nice job generating the data.

projects/graph/src/graph.py
LaikaFusion: I don't believe python allows duplicate keys in a dictionary

projects/graph/src/graph.py
```diff
@@ -5,5 +5,37 @@
...
+        while(len(queue) is not 0):
+            currentKey = queue.pop(0)
+            visited.append(currentKey)
+            if currentKey == searchItem:
+                return True
+            for childkey in self.vertices[currentKey]:
+                if childkey not in visited:
+                    queue.append(childkey)
+        return False
+
```
kelfro: Nice job with the bfs! I think you are right about not being able to find the shortest path with dfs, but it is useful for finding connected components in a graph. In this case, we don't actually need the optimal result, we just need to be able to find one with the available amount of memory: https://cs.stackexchange.com/questions/73686/why-do-we-prefer-dfs-to-find-connected-components

projects/graph/src/graph_demo.py
```diff
@@ -1,16 +1,49 @@
...
     # TODO - parse argv
-    main()
+    # print(argv[1])
+    parser = argparse.ArgumentParser(description='Generate a graph!')
+    parser.add_argument('--vertices', nargs='?', type=int,
+                        help='number of vertices wanted (default is random)')
+    parser.add_argument('--edges', nargs='?', type=int,
+                        help='number of edges wanted (default is random)')
+    args = parser.parse_args()
+    main(args.vertices, args.edges)
```
kelfro: I like the work you did on the argument parser, which is more user friendly than most implementations I've seen. Great job!

projects/graph/src/graph_demo.py
```diff
@@ -37,23 +39,31 @@ def main(numVertices=None, numEdges=None,color=None,style='square',arrows=None):
...
-    parser.add_argument('--arrows',
-                        help='if added will turn on arrows',action="store_true")
+    parser.add_argument('-a', '--arrows',
+                        help='if added will turn on arrows', action="store_true")
+    parser.add_argument('-g', '--groupColor',
+                        help='colors each connection group', action="store_true")
     args = parser.parse_args()
-    main(args.vertices, args.edges,args.color,args.style,args.arrows)
+    main(args.vertices, args.edges, args.color,
+         args.style, args.arrows, args.groupColor)
```
kelfro: This is terrific - I love that you can turn the arrows on and off! Nice work with polishing this so that not only do the color groups and styling work for connected components, but the command line correctly processes arguments for drawing graphs. Well done!

PR #266, opened 2018-10-15
Stephen Fargali

farste:
Initial pull request

### DIFF
--- a/objectives/breadth-first-search/README.md
+++ b/objectives/breadth-first-search/README.md
@@ -82,4 +82,4 @@ BFS(graph, startVert):
 
 * Build a random graph and show a vertex visitation order for BFS.
 
-* One more for good measure.
+* One more for good measure


projects/graph/src/graph.py
```diff
@@ -1,9 +1,82 @@
...
+        for child in children:
+            if child not in visited:
+                child.dft(visited)
+        # For each child, if that child hasn't been visited, call dft() on that node
+        # for child in children:
+        #    if child not in visited:
+                  # dft(child, visted)
+    def bft(self, starting_node):
+        # create an empty queue
+        q = Queue()
```
honda0306: Make sure you define Queue() somewhere before you use it. This is a fine place to have it, just be sure to define what it does too.

objectives/breadth-first-search/bfs.py
```diff
@@ -0,0 +1,17 @@
+from bokeh.plotting import figure, output_file, show
+
+# prepare some data
+x = [1, 2, 3, 4, 5]
+y = [6, 7, 2, 4, 5]
```
honda0306: You've got some good preparation here for BFS; just keep in mind how you set up your data before you start implementing your algorithm.

PR #268, opened 2018-10-15
Charles Boegner

MichaelBoegner:
* [ ] Part 1
* [ ] Part 2
* [ ] Part 3
* [ ] Part 4
* [ ] Part 5
* [ ] Part 6

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Graphs
+# Graphs #
 
 ## Objectives
 


projects/graph/src/graph.py
```diff
@@ -6,4 +6,42 @@
...
+    
+    def get_vertices(self):
+        return [int(k) for k in self.graph]
+        
+    def add_vertex(self, vertex):
+        if vertex not in self.graph: 
+            self.graph[vertex] = {}
+    
+    def add_edge(self, vertex, edge):
+        if edge in self.graph:
```
wvandolah: you'll want to check that both the vert and the edge are in your self.graph

projects/graph/src/graph.py
```diff
@@ -6,4 +6,42 @@
...
+            if self.graph[vertex] == {}:
```
wvandolah: using `set` will prob make your life easier.  You don't have to worry about duplicates and you can just the `.add(edge)`

projects/graph/src/graph.py
```diff
@@ -1,9 +1,78 @@
...
+        verticies = list(self.vertices.keys())
+        q = [verticies[0]]
+        print('first q', q)
+        visited = []
+        
+        while len(q) > 0:
+            nextQ = q.pop()
+            if nextQ not in visited:
+                visited.append(nextQ)
+            q.extend(list(self.vertices[nextQ].edges))
```
wvandolah: nest this under your if statement.

projects/graph/src/graph.py
```diff
@@ -1,9 +1,78 @@
...
+            if child not in visited:
+                  self.dft(child, visited)
+    def bft(self):
+        verticies = list(self.vertices.keys())
+        q = [verticies[0]]
+        print('first q', q)
+        visited = []
+        
+        while len(q) > 0:
+            nextQ = q.pop()
```
wvandolah: a queue should be FIFO(first in first out).  `extend` adds to the end of a list, and `pop()` removes from the end. so this is actually acting like a `dft`.  `pop(0)` would remove from the beginning of the list.

projects/graph/src/draw.py
```diff
@@ -1,14 +1,93 @@
...
+
+        x = []
+        y = []
+        ### start of layout code
+        for vertex_id in node_indices:
+            vertex = graph.vertices[vertex_id]
+            x.append(vertex.x)
+            y.append(vertex.y)
+
+        source = ColumnDataSource(
```
wvandolah: Labels!  Good job deciphering the Bokeh docs.

projects/graph/src/graph.py
```diff
@@ -1,9 +1,78 @@
...
+    #         starting_node = verticies[0]
+    #         print('first starting node', starting_node)
+    #     visited.append(starting_node)
+        
+    #     for child in self.vertices[starting_node].edges:
+    #         if child not in visited:
+    #               self.dft(child, visited)
+    def dbft(self, popper=0):
+        verticies = list(self.vertices.keys())
+        q = [verticies[0]]
```
wvandolah: This is great.  You could pass in a value to your method and use that as your starting point.

projects/graph/src/graph_demo.py
```diff
@@ -5,12 +5,84 @@
...
+    graph.dbft(-1)
+    bg = BokehGraph(graph)
+    bg.draw()
+
+def createRandomGraph(numNodes):
+    graph = Graph()
+
+    edges = []
+    
+    for i in range(numNodes):
```
wvandolah: 👍 this is nice. I see you're passing in edges to your main function, but i don't see it being used anywhere.

PR #271, opened 2018-10-15
Mike Kerbleski

MKerbleski:


projects/graph/src/graph_demo.py
```diff
@@ -8,8 +10,18 @@
 
 
 def main():
-    pass  # TODO
-
+    graph = Graph()
+    graph.add_vertex(0)
```
wvandolah: You could also just use some loops to do all this for you.

projects/graph/src/graph.py
```diff
@@ -1,9 +1,53 @@
...
-        pass  # TODO
+        self.vertices = {}
+    # def add_vertex(self, v):
+    #     self.vertices[v] = set()
+    def add_vertex(self, v):
+        self.vertices[v] = Vertex(v)
+    def add_edge(self, v1, v2):
+        self.vertices[v1].edges.add(v2)
+        self.vertices[v2].edges.add(v1)
+    def dfs(self, value, visited=None):
```
wvandolah: you're on the right track.  a few pointers:
- we don't need to search if a vertex exists, we want to return every vertices that has a connection to the `value` being passed in
- you don't need to loop thru all the vertices every time you call `dfs` you can start with `self.vertices[value].edges`

PR #272, opened 2018-10-15
Alberto Preciado: WIP

Alberto-Funk:


projects/graph/src/graph.py
```diff
@@ -6,4 +6,26 @@
...
+    
+    def add_vertex(self, vert):
+        if not vert in self.verts:
+            self.verts[vert] = set()
+        else:
+            print(f"Vert {vert} already exists")
+    
+    def add_edge(self, vert, edge):
+        if vert in self.verts:
+            self.verts[vert].add(edge)
```
honda0306: Alberto, I like how efficient your implementation is here. The simplicity here will make other parts of your project easier to work with in the coming days.

PR #273, opened 2018-10-15
Jurgen Kela

kelajatu:


projects/graph/Pipfile.lock
```diff
@@ -0,0 +1,130 @@
+{
```
FireInjun: Next time you get more than one pipfile.lock you should delete one so you don't get into a debugging nightmare trying to figure out which `pipenv` you're running. Oftentimes it's something as simple as this that can leave you angry and confused about what's going on in your code.

projects/graph/README.md
```diff
@@ -189,6 +185,6 @@ Highlight the route.
 ### Stretch 5: Travelling saleperson
 
 Read about the [travelling salesperson problem](https://en.wikipedia.org/wiki/Travelling_salesman_problem).
-This is a *hard* problem, literally, so even as a stretch goal a real
+This is a _hard_ problem, literally, so even as a stretch goal a real
 implementation is not expected - but it's an interesting and famous problem
 in the space, and worth learning about if you want to go deep in graphs.
```
FireInjun: Whenever they ask questions in the README you should consider making a new files, something like a *notes.md* and take some time to research the answers. The core concepts can sometimes become more clear if you get stuck somewhere on a particularly difficult piece of code.

projects/graph/src/draw.py
```diff
@@ -1,14 +1,69 @@
 """
```
FireInjun: Consider adding a `shebang` to the top of your Python files. They help to ensure the correct version is executing the file and in some cases they make them easier to work with by making them executable without having to type *python3* in front of it every time in the terminal.

PR #276, opened 2018-10-15
Jun Hyuk Kim: Graphs - In Progress

junhyukee:
- [ ] Part 6: Refactor


projects/.gitignore
```diff
@@ -1,2 +1,3 @@
 .DS_Store
-.vscode/
\ No newline at end of file
+.vscode/
+.idea/
```
kelfro: Did you do this after you already pushed the .vscode and .idea directories? Because if not, it's weird that they are still in here.

projects/graph/src/draw.py
```diff
@@ -2,13 +2,57 @@
...
+        graph.node_renderer.glyph = Circle(size=20, fill_color='color')
+
+        graph.edge_renderer.data_source.data = dict(
+            start=node_indices,
+            end=edges)
+
+        position = {}
+        for vertex in self.graph.vertices:
+            position[vertex] = (random.random() * 4, random.random() * 4)
+
```
kelfro: Great job! The colors and shapes and vertices are fun to play around with - trying out different things to see what happens can also help you learn about bokeh. The documentation is kind of a rabbit hole you can really get lost in.

projects/.gitignore
junhyukee: Noticed that the .gitignore was nested, and i opened the project on the highest level, so it didn't register :( Will do next time!

projects/graph/src/graph.py
```diff
@@ -1,9 +1,74 @@
...
+                    queue.append(next_node)
+                    visited[next_node] = True
+        return res
+
+    def connected(self):
+        visited = []
+        for _, vertex in self.vertices.items():
+            if vertex not in visited:
+                visited.extend(self.dfs(vertex))
+        return visited
```
kelfro: Nice job using dfs for connected components!

projects/graph/src/dijkstra.py
```diff
@@ -0,0 +1,65 @@
...
+graph.add_vertex('0')
+graph.add_vertex('1')
+graph.add_vertex('2')
+graph.add_vertex('3')
+graph.add_edge('0', '1', 3)
+graph.add_edge('0', '2', 5)
+graph.add_edge('1', '3', 7)
+graph.add_edge('2', '3', 1)
+
+print(graph.dijkstra('0', '3'))
```
kelfro: Great job on the stretch problem here! It prints the result in the console - would be terrific to see this plotted in a graph.

kelfro:
Jun, great work on this Graphs project! Your graph displays the colored and styled connected components correctly and is well-written and easy to follow. I also like your thoughts on the travelling salesman problem - it's one of those things that's almost impossible to solve in an efficient way, but good to think about. You clearly put a lot of thought and effort into the basic project and the stretch goals - well done!

PR #278, opened 2018-10-15
Joseph Stossmeister: WIP

Jstoss:
- [ ] graph/vertex/edge
- [ ] drawing with bokeh
- [ ] random graph
- [ ] BFS/DFS
- [ ] Coloring
- [ ] Refactor

### DIFF
--- a/projects/graph/src/graph.py
+++ b/projects/graph/src/graph.py
@@ -1,9 +1,36 @@
 """
 Simple graph implementation compatible with BokehGraph class.
 """
+from random import random
+
+
+class Vertex:
+    def __init__(self, label):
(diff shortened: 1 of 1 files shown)


projects/graph/src/graph.py
```diff
@@ -1,9 +1,56 @@
...
 """
+from random import random
+
+
+class Vertex:
+    def __init__(self, label):
+        self.label = label
+        self.edges = set()
+        self.x = random() * 5
+        self.y = random() * 5
```
honda0306: Nice use of random() here! It's not necessary for vertex generation, but it will give you something new to look at each time.

PR #282, opened 2018-10-15
Justin Hammett | Graphs | Unicorn

hammettDev:


projects/graph/src/graph.py
```diff
@@ -2,8 +2,109 @@
...
-        pass  # TODO
+        self.vertices = {}
+
+    def add_vertex(self, vertex):
+# if key doesn't exsist in the keys we create above then add it to the end of the dictionary. This should just maybe do a "'6': set()" for expample ^set command
+        if vertex.name in self.vertices:
+            raise Exception("ALREADY EXSISTS!")
+
+        self.vertices[vertex.name] = vertex
+        # pass
```
Mister-Corn: As discussed during lecture, an alternative to this is to simply pass a string name for the vertex, and invoke the `Vertex' class' constructor function to make a Vertex object from that. It's up to you if you want to modify your code to accommodate that or keep the course we have now.

projects/graph/src/graph.py
```diff
@@ -2,8 +2,109 @@
...
+
+    def add_edge(self, vert1, vert2):
+        """
+        vert1 and vert 2 are a vertex object
+        gives both vertices a bidirectional edge
+        vert1 and vert2 has an `add_edge` method
+        """
+        vert1.add_edge(vert2)
+        vert2.add_edge(vert1)
+
```
Mister-Corn: Previous comment also applies to this as well:
> As discussed during lecture, an alternative to this is to simply pass a string name for the vertex, and invoke the `Vertex' class' constructor function to make a Vertex object from that. It's up to you if you want to modify your code to accommodate that or keep the course we have now.

projects/graph/src/graph.py
```diff
@@ -2,8 +2,109 @@
...
+# if key exsists then somehow change the values to reflect the connection between the two nodes. ^Add *add to both 
+# counter = -1
+
+class Vertex:
+    def __init__(self, name):
+        self.name = str(name)    # 0
+        self.edges = set()  # a set to all other vertices it's connected to
+
+    def __eq__(self, other):
+        return self.name == str(other)
```
Mister-Corn: Keep the above commented for now. We'll revisit this later if needs be.

projects/graph/src/graph_demo.py
```diff
@@ -14,13 +14,17 @@ def main(num_vertices=8, num_edges=8):
...
+            # print(graph.vertices.keys())
+    # for num in range(num_vertices):
+    #     for _ in range(num_edges):
+    #         print(graph.vertices)
+            # vertices = sample(graph.vertices.keys(), 2)
+            # print(graph.vertices)
             # TODO check if edge already exsists
-            
-            graph.add_edge(vertices[0], vertices[1])
+            # if graph.vertices.values().get_edges() == vertices:
```
Mister-Corn: What are you trying to do? `graph.vertices.values()` should get you something that's like a list, and in it would contain a bunch of vertex objects. Remember: it's a list(_-like thing?_), so it's _not_ going to have the method `get_edges()`. Only the individual vertex objects would have that.

projects/graph/src/graph_demo.py
```diff
@@ -14,13 +14,17 @@ def main(num_vertices=8, num_edges=8):
     for num in range(num_vertices):
         # print(graph.add_vertex(str(num)))
         graph.add_vertex(str(num))
-        print(graph.vertices)
-        for _ in range(num_edges):
-            print(graph.vertices)
-            vertices = sample(graph.vertices.keys(), 2)
```
Mister-Corn: We'll have to play with it together, but trying to establish the edges should probably be a _non-nested_ for loop. In essence, make the vertices with one loop. Then in a totally separate for loop, establish the edges.

PR #283, opened 2018-10-15
Vance Leon

vanceleon:
https://trello.com/b/13LHjTl8/graphs

### DIFF
--- a/LectureNotes.md
+++ b/LectureNotes.md
@@ -1,0 +1,27 @@
+   A B C D E F
+ +-------------
+A | 0 0 1 1 1 0
+B | 0 0 1 0 0 0
+C | 1 0 0 1 0 0
+D | 1 0 1 0 0 0
+E | 1 0 0 0 0 0
+F | 0 0 0 0 0 0
+## Using an Adjacency List
+Another way is to store a list of verts that a particular vert connects
+to.
+A: [ C D E ]
+B: [ C ]
+C: [ A D ]
+D: [ A C ]
+E: [ A ]
+F: []
+Adj Matrix
+Add node: O(N)
+Space complexity: O(N ^ 2)
+Is Connected: O(1)
+Print All Edges: O(N^2), N is number of nodes
+Adj List
+Add node: O(1)
+Space complexity: O(N * e), where e is avg number of edges between nodes
+Is Connected?: O(e), e is avg edges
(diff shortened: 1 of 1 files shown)


projects/graph/src/graph.py
```diff
@@ -6,4 +6,46 @@
...
+        # vert = f"Please specify vertex: {input()}"
+        
+        # self.vertices[vert].update(edge) 
+        # print(self.vertices)
+        if vert not in self.vertices:
+            raise Exception(f"No{vert} vertex")
+            self.vertices[vert].add(edge2)
+            if vert not in self.vertices:
+                raise Exception(f"No {edge2} vertex")
+            self.vertices[vert].add(edge2)
```
Mister-Corn: You might have fixed this already, but there's no need to have three `self.vertices[vert].add(edge2)` here. Your `if..else..` statements can merely check for `vert` in `self.vertices`, and just raise errors as needed. Otherwise, execution will proceed to line 32 and establish the bidirectional edge as normal.

LectureNotes.md
```diff
@@ -0,0 +1,27 @@
...
+Adj Matrix
+Add node: O(N)
+Space complexity: O(N ^ 2)
+Is Connected: O(1)
+Print All Edges: O(N^2), N is number of nodes
+Adj List
+Add node: O(1)
+Space complexity: O(N * e), where e is avg number of edges between nodes
+Is Connected?: O(e), e is avg edges
+Print All Edges: O(E), E is TOTAL number of edges
```
Mister-Corn: Putting notes in your repo is a pretty cool idea. :+1:

projects/graph/src/graph.py
```diff
@@ -1,9 +1,85 @@
...
+    
+    def add_vertex(self, vertex_id):
+        self.vertices[vertex_id] = Vertex(vertex_id)
+
+    def add_edge(self, v1, v2):
+        if v1 in self.vertices and v2 in self.vertices:
+            self.vertices[v1].edges.add(v2)
+            self.vertices[v2].edges.add(v1)
+        else:
+            raise IndexError("That vertex does not exist!")
```
Mister-Corn: We discussed this in DM. Nice work on using Vertex objects. You will find that will be useful when you need to have a place to store `visited` status, x/y coords, color, etc of given vertex somewhere.

PR #285, opened 2018-10-15
Liz Baker 🎉 uNiCoRn 🦄

LizBaker:
- [x] create a graph class
- [ ] bokeh!
- [ ] draw random graph
- [ ] dfs or bfs
- [ ] color
- [ ] refactor

### DIFF
--- a/projects/graph/src/graph.py
+++ b/projects/graph/src/graph.py
@@ -6,4 +6,19 @@ Simple graph implementation compatible with BokehGraph class.
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
-        pass  # TODO
+        self.vertices = {
+            "A": {"B", "C"},
+            "B": {"A"},
+            "C": {"B"}
+        }
+    def add_vertex(self, name):
+        self.vertices[name] = set()
+
+    def add_edge(self, vertex1, vertex2):
+        self.vertices[vertex1].add(vertex2)
+        
+lilgraph = Graph()
+lilgraph.add_vertex("D")
+lilgraph.add_edge("D", "C")
+lilgraph.add_edge("D", "B")
+print(lilgraph.vertices)


projects/graph/src/graph.py
```diff
@@ -6,4 +6,14 @@
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
-        pass  # TODO
+        self.vertices = {
+            0: {1, 2},
+            1: {1},
+            2: {0}
+        }
```
Mister-Corn: If you ever need to initialize an empty set, the syntax is `set()`. You have to use that, because `{}` is reserved for empty dictionaries.

projects/graph/src/graph.py
```diff
@@ -6,4 +6,14 @@
...
+    def add_vertex(self, name):
+        self.vertices[name] = set()
+
+    def add_edge(self, vertex1, vertex2):
+        self.vertices[vertex1].add(vertex2)
```
Mister-Corn: It may be a good idea to allow some option for `add_edge` to create bidirectional edges. For example, if you establish an edge from vertex '0' to vertex '3', it should also establish an edge from vertex '3' to vertex '0' if an option was passed like: `add_edge(a, b, True)` or `add_edge(a, b, bidirectional=True)`.

projects/graph/src/graph.py
```diff
@@ -6,4 +6,14 @@
...
+
```
Mister-Corn: As discussed in lecture, consider implementing the use of vertex objects. In that scenario, `self.vertices` would still be a dictionary, with each key a name of a vertex, but the value of each key would be a vertex object. 

The vertex object itself can have several attributes. For example, you can have an `id` attribute which identifies the vertex, an `edges` attribute which contains a set of other vertices it connects to, etc. etc. You will find more attributes you'd like to add as you progress through the project. The code in lecture is a great starting point.

projects/graph/src/graph.py
```diff
@@ -4,6 +4,54 @@
...
+    def depth_first(self, start, firstrun=True):
+        if firstrun == True:
+            self.answer = []
+        cb = lambda x: self.answer.append(x)
+        cb(start.name)
+        if start.edges:
+            for edge in start.edges:
+                if edge not in self.answer:
+                    self.depth_first(self.vertices[edge], False)
```
Mister-Corn: The use of recursion suggests the use of a stack, as we talked earlier in standup. I will test when I get home, but if you're unsure which search is this, I'd err to "depth-first".

projects/graph/src/graph.py
Mister-Corn: You've implemented a Vertex class. 👍 Don't forget to resolve this convo.

projects/graph/src/graph.py
```diff
@@ -4,6 +4,54 @@
...
 class Graph:
-    """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
-        pass  # TODO
+        self.vertices = {}
+        self.id = -1
+        self.answer = []
+    def createID(self):
+        self.id += 1
+        return self.id
```
Mister-Corn: Oh this is a cool idea! I was trying to implement something similar with Justin, but I couldn't execute. Having it a method in `Graph` is sweet. 😍

PR #290, opened 2018-10-15
Sania Parekh Unicorn CS12 Graphs

sparekh315:


projects/graph/src/graph.py
```diff
@@ -6,4 +6,28 @@
...
-        pass  # TODO
+        self.vertices = {}
+
+    def add_vertex(self, vertex):
+        self.vertices[vertex] = set()
+    
+    def add_edge(self, vertex, dest):
+        self.vertices[vertex].add(dest)
+        self.vertices[dest].add(vertex)
+
```
Mister-Corn: As discussed in lecture, consider implementing the use of vertex objects. In that scenario, `self.vertices` would still be a dictionary, with each key a name of a vertex, but the value of each key would be a vertex object. 

The vertex object itself can have several attributes. For example, you can have an `id` attribute which identifies the vertex, an `edges` attribute which contains a set of other vertices it connects to, etc. etc. You will find more attributes you'd like to add as you progress through the project.

The code in lecture is a great starting point. You may have also noticed example code already in this repo as well to guide you in this implementation.

PR #291, opened 2018-10-15
Richard Verdier - Class defined needs to be tested

rverdi642:
graph.py - done needs testing
draw.py - ToDo
graph_demo.py - ToDo

Researching Bokeh itself

### DIFF
--- a/.idea/markdown-navigator.xml
+++ b/.idea/markdown-navigator.xml
@@ -1,0 +1,82 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="MarkdownProjectSettings" wasCopied="true">
+    <PreviewSettings splitEditorLayout="SPLIT" splitEditorPreview="PREVIEW" useGrayscaleRendering="false" zoomFactor="1.0" maxImageWidth="0" showGitHubPageIfSynced="false" allowBrowsingInPreview="false" synchronizePreviewPosition="true" highlightPreviewType="NONE" highlightFadeOut="5" highlightOnTyping="true" synchronizeSourcePosition="true" verticallyAlignSourceAndPreviewSyncPosition="true" showSearchHighlightsInPreview="false" showSelectionInPreview="true" openRemoteLinks="true" replaceUnicodeEmoji="false" lastLayoutSetsDefault="false">
+      <PanelProvider>
+        <provider providerId="com.vladsch.idea.multimarkdown.editor.swing.html.panel" providerName="Default - Swing" />
+      </PanelProvider>
+    </PreviewSettings>
+    <ParserSettings gitHubSyntaxChange="false" emojiShortcuts="1" emojiImages="0">
+      <PegdownExtensions>
+        <option name="ABBREVIATIONS" value="false" />
+        <option name="ANCHORLINKS" value="true" />
+        <option name="ASIDE" value="false" />
+        <option name="ATXHEADERSPACE" value="true" />
+        <option name="AUTOLINKS" value="true" />
+        <option name="DEFINITIONS" value="false" />
+        <option name="DEFINITION_BREAK_DOUBLE_BLANK_LINE" value="false" />
+        <option name="FENCED_CODE_BLOCKS" value="true" />
+        <option name="FOOTNOTES" value="false" />
+        <option name="HARDWRAPS" value="false" />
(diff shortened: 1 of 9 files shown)


.idea/markdown-navigator.xml
```diff
@@ -0,0 +1,82 @@
...
+      <ScriptProviders />
+      <cssText />
+      <cssUriHistory />
+    </CssSettings>
+    <HtmlExportSettings updateOnSave="false" parentDir="" targetDir="" cssDir="" scriptDir="" plainHtml="false" imageDir="" copyLinkedImages="false" imageUniquifyType="0" targetExt="" useTargetExt="false" noCssNoScripts="false" linkToExportedHtml="true" exportOnSettingsChange="true" regenerateOnProjectOpen="false" linkFormatType="HTTP_ABSOLUTE" />
+    <LinkMapSettings>
+      <textMaps />
+    </LinkMapSettings>
+  </component>
+</project>
```
kelfro: Can you add .idea/ to your .gitignore - I don't think the repo needs to include all this stuff.

projects/graph/src/graph.py
```diff
@@ -5,5 +5,37 @@
...
+        self.points_to = {}
+
+    def addNextedge(self, dest, weight):
+        self.points_to[dest] = weight
+
+    def getNextEdge(self):
+        return self.points_to.keys
+
+    def whoDoes_itPointTo(self,dest):
+        return dest in self.points_to
```
kelfro: This is great that you already started setting up your Vertex class. You're probably going to need it later.

projects/graph/src/graph_demo.py
```diff
@@ -4,13 +4,43 @@
...
+        # graph.add_vertex(Vertex(label=str(num)))
+     graph.add_vertex(Vertex(num))
+    # Add random edges between vertices
+    for _ in range(num_edges):
+       # vertices = sample(graph.vertices.keys(), 2)
+        # TODO check if edge already exists
+        rand_vert = int(random.random() * num_vertices)
+        rand_edge = int(random.random() * num_edges)
+        print(rand_vert, rand_edge)
+        graph.add_edge(vertices[0], vertices[1])
```
kelfro: Would this need to be num_vertices instead of vertices? Let me know if you want help with debugging this. It looks like a great start on getting the random graph to draw but the interpreter is getting hung up on this line.

projects/graph/src/graph.py
```diff
@@ -1,9 +1,75 @@
...
+graph.add_vertex('1')
+graph.add_vertex('2')
+graph.add_vertex('3')
+graph.add_vertex('4')
+graph.add_vertex('5')
+graph.add_vertex('6')
+graph.add_vertex('7')
+# graph.add_edge('0', '1')
+# graph.add_edge('0', '3')
+print(graph.vertices)
```
kelfro: I think because you commented the edges out here, I'm getting empty sets for each vertex in the console and no graph is generated. Maybe we can work on displaying a graph today and getting the dfs and bfs coded because that is going to be on the sprint challenge tomorrow. It would also be good to at least understand how components would be connected using these searches.

projects/graph/src/graph.py
```diff
@@ -1,9 +1,137 @@
...
+        # Otherwise, for each child
+        for child_vert in self.vertices[start_vert].edges:
+            if child_vert not in visited:  # If it hasn't been visited yet
+                # Call dfs_path on the children
+                new_path = self.dfs_path(child_vert, target_value, visited, extended_path)
+                # Return the path if it's valid
+                if new_path:
+                    return new_path
+        return None
+
```
kelfro: Nice work on the searches here! This all looks like it will work to connected components in order of traversal.

projects/graph/src/graph_demo.py
```diff
@@ -4,13 +4,133 @@
...
+    graph.add_vertex(3)
+    graph.add_edge(5, 2)
+    graph.add_edge(5, 6)
+    graph.add_edge(2, 1)
+    graph.add_edge(2, 4)
+    graph.add_edge(4, 3)
+    graph.add_edge(6, 7)
+
+    print(graph.dfs_path(5, 7))
+
```
kelfro: This prints correctly in the terminal, but because it is hard coded in, the same numbers appear connected in the same order - they are just in different locations on the graph.

PR #292, opened 2018-10-15
Will Kwon-Graphs

wtkwon:
- [x] Part 1: Graph, Vertex, Edge Classes

- [ ] Part 6: Refactor

### DIFF
--- a/projects/graph/src/graph.py
+++ b/projects/graph/src/graph.py
@@ -6,4 +6,25 @@ Simple graph implementation compatible with BokehGraph class.
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
-        pass  # TODO
+        self.vertices={}
+    def add_vertex(self,vertex):
+        self.vertices[vertex]=set()
+    def add_edge(self,vertex,edge):
+        try:
+            if self.vertices[vertex] is not None and self.vertices[edge] is not None:
+                self.vertices[vertex].add(edge)
+                self.vertices[edge].add(vertex)
+        except:
+            print('No such vertex exists.')
+        
+
+graph = Graph()  # Instantiate your graph
+graph.add_vertex('0')
+graph.add_vertex('1')
+graph.add_vertex('2')
+graph.add_vertex('3')
+graph.add_edge('0', '1')
+graph.add_edge('0', '3')
+print(graph.vertices)
+graph.add_edge('0', '4')
+print(graph.vertices)


projects/graph/src/graph.py
```diff
@@ -6,4 +6,15 @@
...
-        pass  # TODO
+        self.vertices={}
+    def add_vertex(self,vertex):
+        self.vertices[vertex]=set()
+    def add_edge(self,vertex,edge):
+        try:
+            if self.vertices[edge] is not None:
+                self.vertices[vertex].add(edge)
+                self.vertices[edge].add(vertex)
+        except:
```
wvandolah: https://docs.python.org/3.7/library/exceptions.html#KeyError

projects/graph/src/graph.py
```diff
@@ -2,8 +2,32 @@
...
+    def add_edge(self,v1,v2):
+        if v1 in self.vertices and v2 in self.vertices:
+            self.vertices[v1].edges.add(v2)
+            self.vertices[v2].edges.add(v1)
+        else:
+            raise IndexError('That vertex does not exist.')
+        
+
+class Vertex:
+    def __init__(self,vertex_id,x=None,y=None):
```
wvandolah: You could add self.color to vertex and randomly generate a color for each.  Then override that color when you get to part 5 of this repo.

projects/graph/src/draw.py
```diff
@@ -1,14 +1,48 @@
...
+        for vertex_id in vertex_indices:
+            vertex=graph.vertices[vertex_id]
+            x.append(vertex.x)
+            y.append(vertex.y)
+
+        graph_layout = dict(zip(vertex_indices, zip(x, y)))
+        graph_renderer.layout_provider = StaticLayoutProvider(graph_layout=graph_layout)
+
+        plot.renderers.append(graph_renderer)
+        label_source=ColumnDataSource(data=dict(x=x,y=y,names=[vertex_id for vertex_id in graph.vertices]))
```
wvandolah: 👍

projects/graph/src/graph.py
```diff
@@ -2,8 +2,40 @@
...
+        self.vertices={}
+    def add_vertex(self,vertex_id):
+        self.vertices[vertex_id]=Vertex(vertex_id)
+    def add_edge(self,v1,v2):
+        if v1 in self.vertices and v2 in self.vertices:
+            self.vertices[v1].edges.add(v2)
+            self.vertices[v2].edges.add(v1)
+        else:
+            raise IndexError('That vertex does not exist.')
+    def bfs(self,starting_node):
```
wvandolah: spot on. if you don't want to write a dfs method. just pass in a pop index to this method. `def bdft(self, starting_node, iPop=0):` then when you call the method you have the option of dfs (-1) or bfs(0).

PR #293, opened 2018-10-15
Charlie Winslow

cwinslow22:

- [ ] Part 6: Refactor

### DIFF
--- a/projects/graph/src/graph.py
+++ b/projects/graph/src/graph.py
@@ -6,4 +6,19 @@ Simple graph implementation compatible with BokehGraph class.
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
-        pass  # TODO
+        self.vertices = {}
+    def add_vertex(self, vertex):
+        self.vertices[vertex] = set()
+    def add_edge(self, vert1, vert2):
+        self.vertices[vert1].add(vert2)
+        self.vertices[vert2].add(vert1)
+
+
+graph = Graph()  # Instantiate your graph
+graph.add_vertex('0')
+graph.add_vertex('1')
+graph.add_vertex('2')
+graph.add_vertex('3')
+graph.add_edge('0', '1')
+graph.add_edge('0', '3')
+print(graph.vertices)


projects/graph/src/draw.py
```diff
@@ -1,14 +1,69 @@
...
 General drawing methods for graphs using Bokeh.
 """
-
+import math
 from bokeh.io import show, output_file
 from bokeh.plotting import figure
 from bokeh.models import (GraphRenderer, StaticLayoutProvider, Circle, LabelSet,
-                          ColumnDataSource)
+                          ColumnDataSource, Oval)
+from bokeh.palettes import Spectral8
```
honda0306: I had never heard of this library before, good choice!

projects/graph/src/graph.py
```diff
@@ -6,4 +6,19 @@
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
-        pass  # TODO
+        self.vertices = {}
+    def add_vertex(self, vertex):
+        self.vertices[vertex] = set()
+    def add_edge(self, vert1, vert2):
+        self.vertices[vert1].add(vert2)
+        self.vertices[vert2].add(vert1)
```
honda0306: Very concise implementation of add_vertex and add_edge functions.

projects/graph/src/draw.py
```diff
@@ -28,6 +28,8 @@ def draw(self):
         node_colors = ['red'] * int(N / 2)
         another_color = ['blue'] * int(N/2)
         node_colors.extend(another_color)
+        if N % 2 != 0:
+            node_colors.extend(['green'])
```
honda0306: Good addition for the visual feel of your graphs.

projects/graph/src/graph.py
```diff
@@ -1,9 +1,112 @@
...
+                for edge in self.vertices[current].edges:
+                    q.enqueue(edge)
+
+    def connected(self):
+        color_count = 0
+        colors = ['yellow', 'red', 'blue', 'orange', 'green', 'purple', 'lime', 'olive', 'gold']
+        for vert in self.vertices:
+            if not self.vertices[vert].color != 'grey':
+                self.bft(vert, colors[color_count])
+                color_count +=1
```
honda0306: I think this is the block you're referring to;
`if not self.vertices[vert].color != 'grey'`
I am still looking at it, but I believe there is something about the double negative that is working unusually.

PR #294, opened 2018-10-15
UNICORN - ERIK KIMSEY

ErikKimsey:
### Day 1:

- [x]  Get Bokeh setup and runnning 
- [x]  Create Graph class, with add vertex and edge functionality
- [x] Import Graph into BokehGraph
- [x] Instantiate Graph with BokehGraph and draw


projects/graph/src/graph.py
```diff
@@ -6,4 +6,31 @@
...
+
+    def add_vertex(self, a):
+        self.vertices[a] = set()
+        pass
+
+    def add_edge(self, a, b):
+        if(b not in self.vertices):
+            print(f"Node {b} is not in vertex")
+        else:
+            self.vertices[a].add(b)
```
Mister-Corn: It may be a good idea to allow some option for `add_edge` to create bidirectional edges. For example, if you establish an edge from vertex '0' to vertex '3', it should also establish an edge from vertex '3' to vertex '0' if an option was passed like: `add_edge(a, b, True)` or `add_edge(a, b, bidirectional=True)`.

PR #295, opened 2018-10-15
NAAZ S

noshaikh:


projects/graph/src/graph.py
```diff
@@ -2,8 +2,157 @@
...
+        """create an empty dictionary for the vertices"""
+        self.vertices = {}
+
+    def add_vertex(self, vertex_id):
+        """add a vertex to the graph"""
+        self.vertices[vertex_id] = Vertex(vertex_id)
+
+    def add_edge(self, v1, v2):
+        """add an undirected edge to the graph"""
+        if v1 in self.vertices or v2 in self.vertices:
```
adfaris: In undirected graph both v1 and v2 need to be in self.vertices, so the 'or' should be 'and'.

PR #297, opened 2018-10-15
Zachary Baca

cubsfan46350:


projects/graph/src/draw.py
```diff
@@ -1,14 +1,74 @@
...
 class BokehGraph:
     """Class that takes a graph and exposes drawing methods."""
-    def __init__(self):
-        pass  # TODO
+    def __init__(self, Graph):
+        self.Graph = Graph
+    def draw(self):
+        Graph = self.Graph
+
+N = 8
```
adfaris: this code is used for demo purposes, you will need to modify it so that bokeh will draw from the graph class you defined. Please take a look at https://github.com/br80/Graphs-CS12/blob/master/projects/graph/src/draw.py for reference.

PR #299, opened 2018-10-15
Khoa Nguyen CS12 Graphs

munsu99:


projects/graph/src/draw.py
```diff
@@ -1,14 +1,42 @@
...
+# start of layout code
+circ = [i*2*math.pi/8 for i in node_indices]
+x = [math.cos(i) for i in circ]
+y = [math.sin(i) for i in circ]
+
+graph_layout = dict(zip(node_indices, zip(x, y)))
+graph.layout_provider = StaticLayoutProvider(graph_layout=graph_layout)
+
+plot.renderers.append(graph)
+
```
kelfro: This is a great start! Don't forget to implement a Graph class in graph.py and setup this BokehGraph constructor here to accept Graph as an object. This code you've written here is great and will work as your show method!

projects/graph/src/draw.py
```diff
@@ -1,14 +1,75 @@
...
+graph = Graph()  # Instantiate your graph
+graph.add_vertex('0')
+graph.add_vertex('1')
+graph.add_vertex('2')
+graph.add_vertex('3')
+graph.add_edge('0', '1')
+graph.add_edge('0', '3')
+
+bg = BokehGraph(graph)
+bg.draw()
```
kelfro: I think there's something about using Spectral8 that throws a warning, because I'm getting a message that says color is 8 but index is 4 and that these two numbers have to match up. Just so you know. This is a great template for getting started but you may want to look in the docs for a different color scheme/randomizer.

kelfro:
Nice work getting your graph to render! Please move on to BFS and DFS today as this will be a key part of your sprint challenge tomorrow. Also, it would be very helpful if your commit messages were more frequent, descriptive, and meaningful.

PR #301, opened 2018-10-15
Jennifer Player: Graphs

chainchompa:


projects/graph/src/draw.py
```diff
@@ -1,14 +1,53 @@
...
+graph.add_vertex('2')
+graph.add_vertex('3')
+graph.add_edge('0', '1')
+graph.add_edge('0', '3')
+print(graph.vertices)
 
-class BokehGraph:
-    """Class that takes a graph and exposes drawing methods."""
-    def __init__(self):
-        pass  # TODO
```
kelfro: Your code draws a graph but the BokehGraph class should be implemented with Graph (from graph.py) as an object. I think it is just plotting the vertices and edges above and the code below is not being implemented at all. Is this in progress?

projects/graph/src/graph_demo.py
```diff
@@ -3,14 +3,65 @@
...
+            if arg_split[0] == "style":
+                style = arg_split[1].lower()
+            elif arg_split[0] == "verts":
+                num_verts = int(arg_split[1])
+            elif arg_split[0] == "edges":
+                num_edges = int(arg_split[1])
+    print(style)
+
+    getGraph(style=style, num_verts=num_verts, num_edges=num_edges)
+    getRandomGraph(5, 2)
```
kelfro: When I run graph_demo, I get the same output in my terminal every time, which are the values you hard coded in lines 12-26. But the colors are different each time. Have you tried commenting out the values and seeing if you can generate random numbers for the graph?

projects/graph/src/draw.py
```diff
@@ -1,14 +1,58 @@
...
+        plot.renderers.append(graph_renderer)
+
+        labelSource = ColumnDataSource(data=dict(x=x, y=y, names=[graph.vertices[vertex_id].value for vertex_id in graph.vertices]))
+        labels = LabelSet(x='x', y='y', text='names', level='glyph',
+                     text_align='center', text_baseline='middle', source=labelSource, render_mode='canvas')
+
+        plot.add_layout(labels)
+
+        output_file('graph.html')
+        show(plot)
```
kelfro: Your draw.py does not draw anymore. I think you need to do a BokehGraph(graph).draw() or something like that at the end.

projects/graph/src/graph.py
```diff
@@ -1,9 +1,79 @@
...
+        stack = []
+        stack.append(start)
+        visited = []
+        while len(stack) > 0:
+            current = stack.pop()
+            visited.append(current)
+            for edge in self.vertices[current]:
+                if edge not in visited and edge not in stack:
+                    stack.append(edge)
+        print(visited)
```
kelfro: Great job! When you connect components you will probably need for these to be returns instead of prints.

PR #304, opened 2018-10-15
James McRae

jaymzsocool:


projects/graph/src/graph.py
```diff
@@ -1,9 +1,58 @@
...
+            self.x = random.random() * 10
+        else:
+            self.x = x
+        if y is None:
+            self.y = random.random() * 10
+        else:
+            self.y = y
+    def __repr__(self):
+        return f"{self.edges}git"
+
```
kelfro: I don't think you need the git here on the end of line 49.

projects/graph/src/draw.py
```diff
@@ -1,14 +1,72 @@
...
+graph.add_vertex('1')
+graph.add_vertex('2')
+graph.add_vertex('3')
+graph.add_edge('0', '1')
+graph.add_edge('0', '3')
+print(graph.vertices)
+
+bg = BokehGraph(graph)
+bg.draw()
+
```
kelfro: Great job! The graph draws and returns the correct values for the dictionary. Looking forward to seeing your random graph.

projects/graph/src/graph_demo.py
```diff
@@ -5,12 +5,70 @@
...
+    graph.add_vertex('2')
+    graph.add_vertex('3')
+    graph.add_edge('0', '1')
+    graph.add_edge('0', '3')    
 
-def main():
-    pass  # TODO
+    bg = BokehGraph(graph)
+    bg.draw()
+
```
kelfro: I think because you have lines 12-22 in here, the graph always has the same number of vertices and edges no matter what you enter as arguments when you run this in python. The vertices and edges appear in random locations on the screen, but I cannot generate a graph with, say, 8 vertices and 2 edges or something. This is just something to keep in mind - I think you have the idea on creating a random graph and should not spend too much time fixing this because DFS and BFS are super important and will appear on your sprint challenge.

projects/graph/src/graph.py
```diff
@@ -32,6 +59,15 @@ def add_directed_edge(self, v1, v2):
...
             raise IndexError("That vertex does not exist!")
+    def dft(self, starting_node, visited=None):        
+        if visited is None:            
+            visited = []
+        visited.append(starting_node)  
+        print(starting_node)      
+        for node in self.vertices[starting_node].edges:
+            if node not in visited:
+                self.dft(node, visited)
+        return visited
```
kelfro: Nice job on the DFT here! The next step would be to use this to graph connect components to each other in order of traversal.

projects/graph/src/graph_demo.py
```diff
@@ -66,8 +66,8 @@ def main(style, numNodes, numEdges):
 
 if __name__ == '__main__':
     style = "default"
-    numNodes = 5
-    numEdges = 5
+    numNodes = random.randint(3,7)
+    numEdges = random.randint(3,10)
```
kelfro: This works to make the graph display differently each time. However, since the vertices and edges are hard coded in, there are always the same number of them, connected in the same way - the glyphs just appear in different places.

PR #306, opened 2018-10-15
david pok - graphs

david-pok:
Part 1: Graph, Vertex, Edge Classes
Part 2: Drawing with Bokeh
Part 3: Draw a Random Graph
Part 4: Implement Breadth-First Search (or Depth-First Search)
Part 5: Color the Connected Components Differently
Part 6: Refactor

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,11 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+
+[dev-packages]
+
+[requires]
+python_version = "3.7"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,0 +1,20 @@
+{
+    "_meta": {
+        "hash": {
+            "sha256": "7e7ef69da7248742e869378f8421880cf8f0017f96d94d086813baa518a65489"
+        },
+        "pipfile-spec": 6,
+        "requires": {
+            "python_version": "3.7"
+        },
+        "sources": [
+            {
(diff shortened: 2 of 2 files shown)


projects/graph/src/graph.py
```diff
@@ -1,9 +1,76 @@
...
+    def __repr__(self):
+        return f"{self.edges}"
+
+
+graph = Graph()  # Instantiate your graph
+graph.add_vertex('0')
+graph.add_vertex('1')
+graph.add_vertex('2')
+graph.add_vertex('3')
+print(graph.vertices)
```
kelfro: I don't know if you noticed this, but when you run draw.py, you get two results in the terminal - the first is this line, which shows empty sets for each vertex, because you haven't added the edges yet, and then another below which returns the correct dictionary.

projects/graph/src/graph_demo.py
```diff
@@ -5,12 +5,86 @@
...
+    # O(n)
+    for edge in edges:
+        print(edge)
+        graph.add_edge(edge[0], edge[1])
+
+    print(len(edges))
+
+    bg = BokehGraph(graph)
+    bg.draw()
+
```
kelfro: I  think this would work if you had a draw method in your BokehGraph class (in draw.py).  Since there is no separate method for drawing you would need to tell it the output file and plot manually.

kelfro:
Graph_demo does display a random graph, though I get an error message in the terminal. Good job so far! Be sure to work on BFS and DFS today because this will be on the sprint tomorrow.

PR #307, opened 2018-10-15
Erica Thompson:WIP

EricaThompson:


projects/graph/src/draw.py
```diff
@@ -1,14 +1,137 @@
 """
```
FireInjun: Consider adding a *shebang* at the top of your Python files and leaving more """Docstring""" comments to help with practicing your in code documentation. They're can be very useful and important when you get a career.

projects/graph/src/draw.py
```diff
@@ -1,14 +1,137 @@
...
+# x = [2 * (i // 3) for i in circ]
+# y = [2 * (i % 3) for i in circ]
+
+# graph_layout = dict(zip(node_indices, zip(x, y)))
+# graph_renderer.layout_provider = StaticLayoutProvider(graph_layout=graph_layout)
+
+# plot.renderers.append(graph_renderer)
+
+# output_file("graph.html")
+# show(plot)
```
FireInjun: I really like how you leave your old code in place as comments so you can learn from your previous experiences. It's a very wise practice that will really pay off in the long run as you'll have a record of constantly improving code to compare to your previous work. 💯

projects/graph/src/graph.py
```diff
@@ -3,7 +3,164 @@
...
+            if current_vert == target_value:
+                return True
+            visited.append(current_vert)
+            for vert in self.vertices[current_vert].edges:
+                if vert not in visited:
+                    q.enqueue(vert)
+        
+        return False
+
+
```
FireInjun: Knowing these two searches like the back of your hand will make you a very valuable employee indeed. I would recommend continuing to study and learn about them as much as you can.

projects/graph/src/graph.py
```diff
@@ -3,7 +3,164 @@
...
+                # call dft() on that
+                self.dfs(node, visited)
+        return visited
+
+
+
+
+
+graph = Graph()  # Instantiate your graph
+graph.add_vertex('0')
```
FireInjun: You should go back and see if you can find a way to do this more programmatically. Constant refactoring and improving can only make you better.

PR #308, opened 2018-10-15
Noah Amar [Graphs]

Noah-Amar:


projects/graph/src/graph.py
```diff
@@ -6,4 +6,31 @@
...
+graph.add_vertex('0')
+graph.add_vertex('1')
+graph.add_vertex('2')
+graph.add_vertex('3')
+graph.add_vertex('4')
+graph.add_vertex('5')
+graph.add_edge('1', '4')
+graph.add_edge('1', '5')
+graph.add_edge('3', '4')
+print(graph.vertices)
```
honda0306: Your graph instance looks good; now on to creating the functions that will create the vertices and edges.

PR #309, opened 2018-10-15
[sukhada gholba]Graphs

sukhadagholba:


projects/graph/src/draw.py
```diff
@@ -1,14 +1,50 @@
...
+
+            
+        graph.edge_renderer.data_source.data = dict(start=[0]*N,end=nodes)
+
+        ##start of layout code
+        circ = [i*2*math.pi/len(nodes) for i in range(len(nodes))]
+        x = [math.cos(i) for i in circ]
+        y = [math.sin(i) for i in circ]
+
+
```
kelfro: Nice job! I like that you played around with some of the values a little - it's fun to see what happens when you change things. I'm glad you have bokeh working on your Mac for the rest of this assignment :)

projects/graph/src/draw.py
```diff
@@ -1,14 +1,107 @@
...
+graph.add_vertex('4')
+graph.add_vertex('5')
+graph.add_vertex('6')
+graph.add_undirected_edge('0', '1')
+graph.add_undirected_edge('0', '3')
+graph.add_directed_edge('1', '2')
+graph.add_directed_edge('1', '4')
+graph.add_directed_edge('4', '5')
+graph.add_directed_edge('4', '6')
+graph.breadth_first_search('0')
```
kelfro: This is getting hung up on this line, so maybe comment it out until you get your bfs method written.

projects/graph/src/graph.py
```diff
@@ -2,8 +2,56 @@
...
+        self.edges = set()
+        self.x = x
+        self.y = y
+
+        if x is None:
+            self.x = random.random() * 10 - 5
+        if y is None:
+            self.y = random.random() * 10 - 5
+
+
```
kelfro: I can't run it because of the bfs error in your Graph object, but is this where you are implementing Part 3? I think in the instructions they say to use the graph_demo.py file because we'll need the vertex class for other things later in the project.

sukhadagholba: Thanks for the feedback, I forgot to comment it out yesterday before committing. It has been fixed now.

projects/graph/src/draw.py
sukhadagholba: Thanks for the feedback, I forgot to comment it out before pushing the changes. It has been fixed now.

projects/graph/src/draw.py
```diff
@@ -1,14 +1,99 @@
...
+graph.add_directed_edge('1', '2')
+graph.add_directed_edge('1', '4')
+graph.add_directed_edge('2', '5')
+graph.add_directed_edge('2', '6')
+graph.breadth_first_search('0')
+graph.depth_first_search('0')
+#graph.add_directed_edge('7','4')
+#graph.add_directed_edge('7','4')
+g1 = BokehGraph(graph)
+g1.show()
```
kelfro: Your graph looks beautiful! I can see the results from the DFS in the console and everything looks good. I think next is coloring the components differently! Great work.

projects/graph/src/graph_demo.py
```diff
@@ -3,14 +3,93 @@
...
+    graph.add_directed_edge('2', '6')
+    graph.breadth_first_search('0')
+    graph.depth_first_search('0')
+    graph.dfs_search_value('0','1')
+    #graph.dfs_search_value('0','11')
+    #graph.add_directed_edge('7','4')
+    #graph.add_directed_edge('7','4')
+    g1 = BokehGraph(graph)
+    g1.show()
+
```
kelfro: I am getting three graphs everytime I run graph_demo.py. I think one of them is from your draw.py file. The other two are from this file - one is the default graph, and the other is the random graph. I think this is because you run the show() method for each one.

PR #311, opened 2018-10-16
Arthur Bates

AliasAJ:
forgotten pull

### DIFF
--- a/projects/graph/src/draw.py
+++ b/projects/graph/src/draw.py
@@ -7,8 +7,20 @@ from bokeh.plotting import figure
 from bokeh.models import (GraphRenderer, StaticLayoutProvider, Circle, LabelSet,
                           ColumnDataSource)
 
+x = [1, 2, 3, 4, 5]
+y = [6, 7, 2, 4, 5]
 
 class BokehGraph:
     """Class that takes a graph and exposes drawing methods."""
     def __init__(self):
-        pass  # TODO
+        # output to static HTML file
+        output_file("graph1.html")
+
+# create a new plot with a title and axis labels
+        p = figure(title="simple line example", x_axis_label='x', y_axis_label='y')
+
+# add a line renderer with legend and line thickness
+        p.line(x, y, legend="Temp.", line_width=2)
+
+# show the results
+        show(p)
--- a/projects/graph/src/graph.py
+++ b/projects/graph/src/graph.py
@@ -2,20 +2,25 @@
 Simple graph implementation compatible with BokehGraph class.
 """
 
-from collections import defaultdict
-
-
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
-      self.vertices = defaultdict(set)
+        self.vertices = set()
 
-     def add_vertex(self, vertex):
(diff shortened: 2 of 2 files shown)


projects/graph/src/graph.py
```diff
@@ -2,8 +2,106 @@
...
+        if not set(edges).issubset(self.vertices):
+            raise Exception("Error: Cannot have edge to nonexistent vertices")
+        if vertex in self.vertices:
+            raise Exception("Error: Vertex already exists.")
+        self.vertices[vertex] = set(edges)
+
+
+def main():
+    graph = Graph()
+    graph.add_vertex("0")
```
FireInjun: You should go back and see if you can do this in a more programmatic way to challenge yourself.

projects/graph/src/graph.py
```diff
@@ -2,8 +2,106 @@
...
+            visited = []
+        visited.append(starting_node)
+        # For each child, if that child hasn't been visited, call dft() on that node
+        for node in self.vertices[starting_node].edges:
+            if node not in visited:
+                self.dft(node, visited)
+        return visited   
+
+    """
+    BFS(graph, startVert):
```
FireInjun: Multi-line comments are a great way to keep track of what you're doing and document what your code is up to. This is a great habit to have. 👍

projects/graph/src/draw.py
```diff
@@ -2,13 +2,172 @@
...
+        # Setup plot
+        self.width = width
+        self.height = height
+        self.pos = {}  # dict to map vertices by x, y positions
+        # new plot with title and ranges
+        """The bokeh.plotting interface provides a Figure class to help with assembling all the necessary objects.
+            the figure() function creates Figure objects."""
+        self.plot = figure(
+            title=title,
+            # Ranges describe the data-space bounds of a plot, in this case, from 0 to the graphs width and height
```
FireInjun: Remember Python doesn't like long lines or bad whitespace. In fact the official PEP8 documentation says that lines should be no longer that 88 characters. I would consider refactoring to be more Pythonic, it will definitely help you dive deeper into the language and improve your skills. It's always better to see what best practices are for each language.

projects/graph/src/graph.py
```diff
@@ -2,8 +2,106 @@
...
+    for v of graph.verts:
+        v.color = white
+        v.parent = null
+
+    for v of graph.verts:
+        if v.color == white:
+            DFS_visit(v)
+    """
+
+    def dft(self, starting_node, visited=None):
```
FireInjun: This was the hardest part of the week but, it's also the single most important part of what we've studied so far in the CS portion of the curriculum. I would highly recommend you go back and study this until it's ingrained in your mind just as much as create-react-app is. It will mean the difference between getting a junior position or a senior position later on in your career.

projects/graph/src/draw.py
```diff
@@ -2,13 +2,172 @@
 General drawing methods for graphs using Bokeh.
```
FireInjun: Consider adding *shebangs* to the top of each Python file, just to get in the habit of it for when you get a job. They may seem like they aren't that important but they are a central part of the Python methodology.

PR #313, opened 2018-10-17
[Sam Khaled] finished

AppMkrATL:


projects/graph/src/draw.py
```diff
@@ -10,5 +12,107 @@
...
+        self.plot.renderers.append(graph_renderer)
+
+    def _get_random_colors(self, num_colors=None):
+        colors = []
+        num_colors = num_colors or len(self.graph.vertices)
+        for _ in range(num_colors):
+            color = '#'+''.join([choice('0123456789ABCDEF') for j in range(6)])
+            colors.append(color)
+        return colors
+
```
FireInjun: Consider challenging yourself more and hit some stretch goals. I'd like to see what you come up with! 👍

PR #314, opened 2018-11-26
Timothy Hoang - Graphs

timOTdev:
**Day 1 and 2**
- [x] Part 1: Graph, Vertex, Edge Classes + Stretch
- [x] Part 2: Implement Breadth-First Traversal
- [x] Part 3: Implement Depth-First Traversal with a Stack
- [x] Part 3.5: Implement Depth-First Traversal using Recursion
- [x] Part 4: Implement Breadth-First Search
- [x] Part 5: Implement Depth-First Search

**Day 3 and 4**
- Part 6: Write an algorithm that will let you create any number of random adventure rooms.
- [x] Write an algorithm, createRandomRooms that takes in a number and generates that number of rooms.
- [x] Each room should be connected and reachable from the player's starting room.
- [ ] None of the rooms should overlap and they should all exist on a valid grid.
- [x] Some information about the room, like possible exits, should be present in the room description.
- [ ] Try to keep your runtime complexity to O(n).

- Part 7: Randomly drop a Treasure, then write a search method to find it
- [x] Create a large number of rooms (say, 100) then drop a treasure in a random room before the game starts.
- [x] Write a function for your player that uses BFS to find the shortest path to the treasure.


projects/graph/src/graph.py
```diff
@@ -5,5 +5,6 @@
 
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
+
     def __init__(self):
-        pass  # TODO
+        self.graph = {}
```
TheDeterminator: Interesting, why the choice to name this attribute self.graph?

TheDeterminator:
Tim, great job as usual diagramming your workflow. I think you may have a mismatch between your checklist and your pushed code though, I will continue to check back for more pushed commits.

projects/graph/src/graph.py
timh1203: I didn't read instructions clearly yet and just wanted to push up an initial commit.

It has been changed since. 😄

projects/graph/src/graph.py
```diff
@@ -5,5 +5,35 @@
...
+            return False
+
+    def add_edge(self, v1, v2):
+        if v1 in self.vertices and v2 in self.vertices:
+            self.vertices[v1].add(v2)
+            self.vertices[v2].add(v1)
+            return True
+        else:
+            if v1 not in self.vertices:
+                raise Exception(f"Vertex {v1} is nonexistant!")
```
TheDeterminator: This is good! If you want to be more specific you could also do ```raise IndexError``` since we're discussing a non-existant entry into a dict.

TheDeterminator: Also love that you're still using the string formatting that we learned about a few weeks ago.

projects/graph/src/graph.py
```diff
@@ -3,7 +3,124 @@
...
+        s.append(int(target))
+
+        if str(target) not in self.vertices:
+            raise IndexError(f"Vertex {target} is nonexistent!")
+
+        while len(s) > 0:
+            discovered = s.pop()
+            if discovered not in visited:
+                visited.append(discovered)
+                for c in self.vertices[str(discovered)]:
```
TheDeterminator: Is this targeting the edges of each child vertex or is it simply targeting every vertex as they appear in the dict? I'm curious if there are test cases related to the ordering of the vertices (in the dict and when generating edges) that would break this method as is. I will run some test when I get a chance. Having said that the code looks great!

TheDeterminator:
Tim, your code looks great! As usual it is very easy for me to see how you're progressing with your checklists. Your BFT and DFT implementations look good so far, I made a comment on one but it is relatively minor. Only thing I want to mention, as a matter of style in convention it is considered best practice to use more descriptive variable names rather than single character variable names (such as c and s in your DFT method). It's a small thing though. Overall awesome job today.

projects/graph/src/graph.py
timh1203: I am using the sets from the dictionaries and can see why it works.

After talking about it in code review and seeing Brady inserting the dictionary value as Vertex(value) this morning, I realized my implementation is different.

I do see the difference in what you're pointing out targeting off the edges in the class instead of off the edges in the dictionary. Much appreciated!

TheDeterminator:
Tim, your code looks great. Only major thing I'd point out is that I didn't see any loops or recursive calls in your generateRooms function in world.py It makes me wonder how you will go about choosing a new room to try to connect in the event of the collision. I know you mentioned trying to implement a recursive solution but I don't see any recursion in this code. I may have missed it though. All the same great job this week.

timh1203:
> Tim, your code looks great. Only major thing I'd point out is that I didn't see any loops or recursive calls in your generateRooms function in world.py It makes me wonder how you will go about choosing a new room to try to connect in the event of the collision. I know you mentioned trying to implement a recursive solution but I don't see any recursion in this code. I may have missed it though. All the same great job this week.

Yes, I didn't commit anything because it didn't work. Basically, I was attempting to re-run generateRooms function in the else statement, if the coordinate-pair was not in the avoid_list basically.

```
if (self.x_counter, self.y_counter) not in self.avoid_list:
                        self.rooms[i - 1].connectRooms(path, self.rooms[i])
                        self.avoid_list.add((self.x_counter, self.y_counter))
                        print(
                            path,
                            (self.x_counter, self.y_counter),
                            self.avoid_list,
                            f"({len(self.avoid_list)}/{numRooms -1})",
                            f"{(numRooms -1) - (len(self.avoid_list))}",
                        )
else:
                        print("collision")
                        self.generateRooms(numRooms, needed_rooms)
                        print(numRooms - 1, len(self.avoid_list), needed_rooms)
                        break
```

PR #316, opened 2018-11-26
Kevin Nguyen CS13: Graphs

Kevnien:


adrianadames:
Great start Kevin. You're well on your way. Only thing I caught so far is that when I added an edge between vertex label '2' and vertex label '3', the dictionary graph.vertices only shows that vertex label '2' is connected to '3' and not vice-versa. This would make sense if the edge was intended to be a directed edge from '2' to '3'. 

One thing to also consider is making the vertices their own class object with a couple of attributes to start off - a unique id (or label as you have it) and an edges list (i.e. an adjacent vertices list). This might help bring more clarity to your code logic and operations as you progress to BFT and DFT, and leaves the window open to add other possibly useful attributes to your vertices like color and value. Your approach works perfectly fine though.

adrianadames:
Great job Kevin. I tested your code and everything checks out. You might want to return a list for the bft/dft traversals like you do with the bfs/dfs methods. It's not required per the problem statement, but it can be a useful thing to be able to access the full bft/dft traversal of a graph.

PR #317, opened 2018-11-26
Jeremy Boggs - Graphs

Nicocchi:


TheDeterminator:
Jeremy, awesome job so far. Very interesting that you used the built-in python queue module. You're the first person I've seen do that so far. I'll be interested to see what you do with DFS and/or what you do next. Keep it up!

projects/graph/src/graph.py
```diff
@@ -2,52 +2,270 @@
...
+                print(node)
+                visited.add(node)
+                # then put all it's children in the stack
+                for child in self.vertices[node].edges:
+                    s.push(child)
+
+    # I could probably simplify this more
+    def dft_r(self, node, visited=None, s=None):
+        if visited is None:
+            visited = set()
```
TheDeterminator: It's a small thing, but you could save yourself two lines by making the default value of visited an empty list

TheDeterminator:
Jeremy, your code looks awesome as usual. I just want to bring one thing to your attention. Your getRandomDirection function on line 10 of your world.py uses the self.x_to attributes of the room class to test for collisions. This is good up to a point, but it there are some cirumcstances in which using those attributes fails to detect adjacent rooms. Consider this set of rooms:

[_]-[_]
 |   |
[_] [_]
 |   |
[_] [_]
 |   |
[_] [_X_]

The room that is marked as X has no self.w_to() attribute but there is still a room next to that that would cause a collision. That's the pitfall of using self.x_to as opposed to coordinates. Even still great job.

PR #321, opened 2018-11-26
Michelle Paredes - Graphs Project

mparedes003:


TheDeterminator:
Michelle, awesome job, I'm very happy with your work today. As a next step you might consider making a Vertex class for your vertices so that you can store more (and more dynamic) data in your vertex, with your current implementation, the vertex is basically a number string, but you will eventually want to be able to add other values and properties to the vertex. You should also begin thinking about how you will implement BFS keeping in mind that we did a very similar BFS problem in our data structures project.

TheDeterminator:
Michelle, what you have so far is not bad. It could be tough to implement proper collision checking though with only using the x_to methods on the room class. I'm sure you will factor the lecture the info from the lecture into your updated code though.

PR #323, opened 2018-11-26
Arkadiusz Kusiak--Graphs

akusiak23:


projects/graph/src/graph.py
```diff
@@ -5,5 +5,43 @@
```
ari7946: Nice work. The add vertex and add edges methods are working correctly

projects/graph/src/graph.py
```diff
@@ -5,5 +5,43 @@
...
+        else:
+            self.vertices[value] = set()
+
+    def add_edge(self, v1, v2):
+        if v1 not in self.vertices or v2 not in self.vertices:
+            raise Exception('Invalid input: vertices not in graph')
+        else:
+            self.vertices[v1].add(v2)
+            self.vertices[v2].add(v1)
+
```
ari7946: I believe this is dft instead of dfs

PR #324, opened 2018-11-26
GRAPHs Joseph Chretien

JGolden25:
@ari7946


projects/graph/src/graph.py
```diff
@@ -2,8 +2,22 @@
 Simple graph implementation compatible with BokehGraph class.
 """
```
ari7946: Everything looks great so far.

PR #325, opened 2018-11-26
Adam Reid - [Graphs]

iAmAdamReid:


iAmAdamReid:
Current issue: BFsearch seems to be visiting the '0' vertex twice? Not sure how that is happening.

iAmAdamReid:
NOTE: There is some inconsistency with the returned datatypes which could be problematic if using this with external callbacks (some methods return objects and others strings). I will see if I can achieve data consistency on all method returns.

.vscode/tags
```diff
@@ -0,0 +1,56 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
```
TheDeterminator: I would add this to your .gitignore if it's getting pushed in with your commits...alternatively, I wonder if you do ```git add .``` when you add your changes. That's okay, but it adds the "."-prepended files to your commits too. You should use ```git add *``` to add all changed files except for dot-prepended ones.

TheDeterminator:
Adam, your code so far looks stellar. You really grasp everything that we go over very very well. The code is also very clean and well commented. I see you wrote a comment about having issues with returned data types, I'll be interested to see how you handle that issue, I may have some ideas for you if you don't figure it out on your own first. Also have you did you fix that issue with duplicate values being returned from one of your methods? I haven't run your code just yet but I don't see any indications so far that that would still be an issue. Great work today.

projects/graph/adv/world.py
```diff
@@ -5,24 +5,39 @@ class World:
     def __init__(self):
         self.startingRoom = None
         self.rooms = {}
+        self.occupied = set()
 
-    def getRandomDirection(self, room):
+    def getRandomDirection(self, room, coords):
         dirs = []
-        if room.n_to is None:
+        if room.n_to is None and self._checkCoordinates(coords, 'n'):
```
TheDeterminator: This is good, consider though that the room.x_to part may not be super useful. There are circumstances where room.x_to can be undefined but there is a still an adjacent room in the x direction. Consider this layout:

[_]-[_]
 |   |
[_] [_]
 |     |
[_] [_]
 |     |
[_] [X]

I suspect the formatting may get messed up when I submit this comment, but basically the room that I marked as X has a room to the west but it's not "connected" so room.w_to would be None even though there is an adjacent room. That's why using coordinates to check is preferred.

TheDeterminator:
Adam your code looks awesome so far. You did an excellent job with this project. I noticed you made a comment in your EOD report about having a weird room generation in your world.py file. I haven't gotten to look at this too closely yet but I will take some time to do that or else we can talk about it during the stand up meeting. Unless you want to figure it out on your own. Again, great work today and this week!

PR #326, opened 2018-11-26
Brian

BrianARuff:


projects/graph/src/graph.py
```diff
@@ -3,7 +3,36 @@
 """
 
 
+class Vertex:
+    def __init__(self, vertex):
+        self.vertex = vertex
+        self.edges = []
```
ari7946: Would it make sense to use a set instead of a list to prevent adding an edge that is already there?

projects/graph/src/graph.py
```diff
@@ -3,7 +3,113 @@
...
+
+class Vertex:
+    def __init__(self, vertex):
+        self.vertex = vertex
+        self.edges = set()
+
+    def __repr__(self):
+        return f"Edges: {self.edges}, V: {self.vertex}"
+
+
```
ari7946: There's a small space there before the class in case it matters.

projects/graph/src/graph.py
```diff
@@ -3,7 +3,113 @@
 """
```
ari7946: Great job. Everything seems to be working as it should.

PR #327, opened 2018-11-26
Jcassidy

cassidyjamesw:


Dilstom:
Good start, James. Looking forward to seeing your implementation of bfs and dfs.

Dilstom:
Probably, you want to do something with this bft - maybe, print each visited node...

Dilstom:
I like your thinking, you could also use a Stack with this dft. In addition, you could print the resulting nodes.

PR #328, opened 2018-11-26
Michelle Ok CS13: Graphs

Mokagbue:
made a notes file

### DIFF
--- a/Notes.txt
+++ b/Notes.txt
@@ -1,0 +1,1 @@
+GRAPH NOTES


projects/graph/src/graph.py
```diff
@@ -8,8 +8,8 @@ def __init__(self):
         pass  # TODO
         self.vertices = {}
     
-    def add_vertex(self, vertex_id):
-        self.vertices[vertex_id] = Vertex(vertex_id)
+    def add_vertex(self, vertex_id, vertices):
```
adrianadames: I'm a little unclear on this part. I don't think we need the second argument, vertices, since we can assume that the vertex being added is being added to the graph's own vertices dictionary attribute.

projects/graph/src/graph.py
```diff
@@ -8,8 +8,8 @@ def __init__(self):
...
+        self.vertices[vertex_id] = vertices(vertex_id)
```
adrianadames: This line right here is saying, in the dictionary vertices, the key value corresponding to the key 'vertex_id' is 'vertices(vertex_id)'. I'm not sure what this key value 'vertices(vertex_id)' is supposed to return.

adrianadames:
Good first attempt Michelle!

projects/graph/src/graph.py
```diff
@@ -2,28 +2,170 @@
...
+            self.vertices[v1].edges.add(v2)
+            self.vertices[v2].edges.add(v1)
         else:
             raise IndexError("That vertex does not exist!")
 
-    def add_directed_edge(self, vert1, vert2):
-        if vert1 in self.vertices:
-            self.vertices[vert1].edges.add(vert2)
+    def add_directed_edge(self, v1, v2):
+        if v1 in self.vertices:
```
adrianadames: In this part here, you also want to make sure that v2 is in self.vertices.

PR #329, opened 2018-11-26
Jonathan Miles -- Graphs

jonathantmiles:


projects/graph/src/graph.py
```diff
@@ -6,4 +6,14 @@
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
-        pass  # TODO
+        vertices = []
```
TheDeterminator: You'll want to make this a dict. Refer to the README

TheDeterminator:
Jonathan, you have a very creative implementation so far. I'm not sure if I understand your add_edge implementation yet, but I believe it can work. That being said you may want to consider creating a Vertex class to hold data and edges for your vertices. Good job so far.

projects/graph/src/graph.py
```diff
@@ -9,11 +9,15 @@ def __init__(self):
...
     def add_edge(self, vertex_one, vertex_two):
         first_index = vertices.index(vertex_one)
         vertices[first_index] = {vertex_one, vertex_two}
         second_index = vertices.index(vertex_two)
         vertices[second_index] = {vertex_two, vertex_one}
+
+class Vertex:
+    def __init__(self, vertex, edge=set()):
+        self.vertex = vertex
+        self.edge = edge
```
TheDeterminator: A small thing, but it might be nice to call this "edges" instead of simply "edge"

TheDeterminator:
Jonathan, I see you took my advice about adding a Vertex class, it looks very good. Don't forget that you can use the code in the README to test if your implementations are working. I look forward to seeing how you implement DFT and BFT. Let me know if you have any questions.

PR #332, opened 2018-11-26
Graphs - Lucas Beemer

lucasbeemer:
Made Graph class with add edge and add vertex methods. 
Also created Vertex class.

### DIFF
--- a/projects/graph/src/graph.py
+++ b/projects/graph/src/graph.py
@@ -3,7 +3,45 @@ Simple graph implementation compatible with BokehGraph class.
 """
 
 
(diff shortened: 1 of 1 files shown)


projects/graph/src/graph.py
```diff
@@ -128,4 +126,4 @@ def dftr(self, starting_node, visited=[]):
 graph2.add_vertex("5")
 graph2.add_vertex("6")
 graph2.add_vertex("7")
```
ari7946: I think you may need to add some edges too

PR #336, opened 2018-11-26
Huthman King - Graphs

kinghuthman:


projects/graph/src/graph.py
```diff
@@ -6,4 +6,20 @@
 class Graph:
```
ari7946: Everything looks great so far.

projects/graph/src/graph.py
```diff
@@ -2,8 +2,83 @@
...
+        self.vertices={}
+    def add_vertex(self,vertex_id):
+        self.vertices[vertex_id]=Vertex(vertex_id)
+    def add_edge(self,v1,v2):
+        if v1 in self.vertices and v2 in self.vertices:
+            self.vertices[v1].edges.add(v2)
+            self.vertices[v2].edges.add(v1)
+        else:
+            raise IndexError('That vertex does not exist.')
+
```
ari7946: I think these should be dft methods if I'm not mistaken. They do look correct though.

projects/graph/src/graph.py
```diff
@@ -2,8 +2,83 @@
...
+        while queue.len() > 0:
+            vertex = queue.dequeue()
+            if vertex not in visited:            
+                visited.append(vertex)
+                for child in self.vertices.get(vertex):
+                    queue.enqueue(child)
+        return visited
+    
+    def dfs(self, start_vertex):
+        stack = []
```
ari7946: why not use the Stack class you created? I think it has better performance.

ari7946: I think using the stack class you created will have better performance than a list.

PR #339, opened 2018-11-26
finished objective 1: graph, vertex, edge classes

ThomasKisner:


Dilstom:
nice start, Thomas. Now it is time to implement bfs and dfs.

PR #346, opened 2018-11-27
completed mvp for the day--Zachary Baca

cubsfan46350:


projects/graph/src/graph.py
```diff
@@ -1,9 +1,59 @@
...
+            raise IndexError("That vertex does not exist!")
+    def bft(self, starting_node):
+        q = Queue()
+        q.enqueue(starting_node)
+        visited = []
+        while q.size() > 0:
+            visited_item = q.pop(0)
+            if visited_item not in visited:
+                visited.append(visited_item)
+                q.extend(list(visited_item.add_edge))
```
TheDeterminator: This line is interesting, why the choice to use extend  and visited_item.add_edge? It seems that it would be better to use the enqueue function of your Queue class, and it's not clear to me that .add_edge() is appropriate here, maybe I'm misunderstanding?

PR #349, opened 2018-11-27
James McRae

jaymzsocool:


adrianadames:
Very well done James. Your BFS return path is almost there!

adrianadames:
I was able to play your game and move throughout all the different rooms. I don't think it's required, but it'd be nice if the player knew the possible directions he/she could go when they enter a new room. Keep up the good work James!

PR #350, opened 2018-11-28
Luisan Suarez - CS 13

LuisanSuarez:
@dilstom
@luisansuarez


Dilstom:
Great job, Luisan! 💯 How about implementing Depth-First Traversal using Recursion? :)

PR #362, opened 2019-01-28
Troy Amlee initial commit

troyAmlee:
Graph implementation

### DIFF
--- a/projects/graph/src/graph.py
+++ b/projects/graph/src/graph.py
@@ -1,9 +1,49 @@
 """
 Simple graph implementation
 """
-
+"""
+{
+'0': {'1', '3'},
+'1': {'0'},
+'2': {set()},
+'3': {'0'}
+}
+"""
+import pprint
 
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
-    def __init__(self):
-        pass  # TODO
+    def __init__(self, destination = None, label = None):
+        self.vertices = {}
+        self.edges = set()
+        self.destination = destination
+        self.label = label
+
+
+
+    def add_vertex(self, vertex):
+        self.vertices[vertex] = set()
+        print(self.vertices)
+
+    def add_edge(self, vertex_start, vertex_end):
+        for i in self.vertices:
+            # temp = set()
+            # temp.add(i)
+            if i == vertex_start:
+                self.vertices[vertex_start].add(vertex_end)
+                print(self.vertices)
+            if vertex_end == i:
+                self.vertices[vertex_end].add(vertex_start)
+                print(self.vertices)
+
+
+def print_vertex():
+    graph = Graph()
+    graph.add_vertex('0')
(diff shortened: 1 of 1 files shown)


projects/graph/src/graph.py
```diff
@@ -1,9 +1,168 @@
...
+            for i in self.vertices[deq]:
+                if i not in visited:
+                    q.append(i)
+                if len(q) == 0:
+                    return visited
+
+
+    def depth_first_traverse(self, starting_vertex):
+        # create a stack LIFO
+        s = queue.LifoQueue()
```
blauente: You can create your own Queue class in another file or at the top of this file here. This might be easier than importing from the built-in module.

projects/graph/src/graph.py
```diff
@@ -1,9 +1,168 @@
...
+            # dequeue a node from the queue
+            # Mark it as visited
+            # If node == target node: return True
+            # Enqueue all of its children
+        # return False
+
+    def breadth_first_search(self, starting_vertex, target):
+        # create a _queue_ FIFO
+        q = Queue()
+        visited = set()
```
blauente: I think it makes more sense to make every `visited` a list instead of a set because lists are ordered and sets are not.

projects/graph/src/graph.py
```diff
@@ -1,9 +1,168 @@
...
+        # print(starting_vertex)
+        # visited.add(starting_vertex)
+        # Enqueue the starting vertex
+        q.Enqueue(starting_vertex)
+        # while the queue is not empty
+            # dequeue a node from the queue
+            # Mark it as visited
+            # If node == target node: return True
+            # Enqueue all of its children
+        # return False
```
blauente: Please continue to work on BFS and DFS today.

projects/graph/social/social.py
```diff
@@ -0,0 +1,127 @@
...
+
+if __name__ == '__main__':
+    sg = SocialGraph()
+    sg.populateGraph(10, 3)
+    print(sg.friendships)
+    connections = sg.getAllSocialPaths(1)
+    print(connections)
+    sg.populateGraph(100, 10)
+    print(sg.friendships)
+    sg.populateGraph(1000, 5)
```
blauente: It's really great that you tested these methods with higher numbers. Awesome job! 👍

projects/graph/social/graph.py
```diff
@@ -0,0 +1,247 @@
...
+            return
+        if vertex_end not in self.vertices:
+            print(f"You chose: {vertex_end} That End Vertex does not exist...")
+            return
+        for i in self.vertices:
+            # temp = set()
+            # temp.add(i)
+            if i == vertex_start:
+                self.vertices[vertex_start].add(vertex_end)
+                print(self.vertices)
```
blauente: I like that you mix unidirectional and bidirectional edges in the same graph. Very cool.

PR #365, opened 2019-01-28
Graphs-Caleb-Kirkwood-(Team-Grant)

ckirkwood94:
@blauente

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,13 @@
 
 ## Objectives
 
-* [Graph Intro](objectives/graph-intro)
-* [Graph Representations](objectives/graph-representations)
-* [Randomness](objectives/randomness)
-* [Breadth First Search](objectives/breadth-first-search)
-* [Depth First Search](objectives/depth-first-search)
-* [Connected Components](objectives/connected-components)
+- [Graph Intro](objectives/graph-intro)
+- [Graph Representations](objectives/graph-representations)
+- [Randomness](objectives/randomness)
+- [Breadth First Search](objectives/breadth-first-search)
+- [Depth First Search](objectives/depth-first-search)
+- [Connected Components](objectives/connected-components)
 
 ## Projects
 
-* [Random Social Network](projects/graph)
+- [Random Social Network](projects/graph)


projects/graph/src/graph.py
```diff
@@ -1,9 +1,83 @@
...
+            if vertex1 not in self.vertices:
+                undefined.append(vertex1)
+            if vertex2 not in self.vertices:
+                undefined.append(vertex2)
+            undefined = ", ".join(undefined)
+            raise KeyError(f"Vertex not found: {undefined}")
+
+    def breadth_first_traversal(self, starting_node):
+        queue = Queue()
+        visited = set()
```
blauente: I think it would make sense to use lists instead of sets in all of the places where you have `visited` because lists are ordered and sets are not.

projects/graph/src/graph.py
```diff
@@ -1,9 +1,83 @@
...
+graph.add_edge('0', '1')
+graph.add_edge('0', '3')
+graph.add_edge('3', '4')
+graph.add_edge('1', '2')
+graph.add_edge('3', '5')
+graph.add_edge('6', '4')
+
+
+# graph.breadth_first_traversal('0')
+graph.depth_first_traversal_recursive('0')
```
blauente: The vertices look like they're printing correctly when you run `graph_demo.py`. Very nice. 👍

projects/graph/src/graph.py
ckirkwood94: I have it as a set, because when we're checking if we visited a vertex or not it shouldn't matter if the set is in order. We're checking if the vertex is a member of visited to determine if we need to visit or not. Which I believe a set is faster than a list when checking for a value inside.

blauente: Okay, thanks for letting me know. Good answer. 👍

projects/graph/social/social.py
```diff
@@ -45,10 +52,26 @@ def populateGraph(self, numUsers, avgFriendships):
...
+        # for num in range((numUsers*avgFriendships)//2):
+        #     friendship = possible_friendships[num]
+        #     self.addFriendship(friendship[0], friendship[1])
+
+        # Refactor for adding friendships stretch
+        numFriends = 0
+        while numFriends < numUsers*avgFriendships//2:
+            friends_added = self.addFriendship(
+                randint(1, numUsers), randint(1, 10))
+            numFriends += friends_added
```
blauente: You might want to add a check to make sure that the numbers you're passing into `addFriendship()` are not the same, I'm getting some errors of `WARNING: You cannot be friends with yourself`. And if you use for loops you might be able to avoid `WARNING: Friendship already exists` errors.

PR #367, opened 2019-01-28
Graphs--[Lajawanti]

innovationcode:


projects/graph/src/graph.py
```diff
@@ -1,9 +1,48 @@
...
+
+    def add_edge(self, start_point, end_point):
+        if start_point in self.vertices and end_point in self.vertices:
+            #graph is bidirectional.. should add edges back and forth
+            self.vertices[start_point].add(end_point)
+            self.vertices[end_point].add(start_point) 
+        return "vertices not in graph"
+    
+    
+    def breadth_first_traversal(self):
```
blauente: I think the way most people have this set up is that you pass in the node/vertex you want to start at, so you can test different starting points.

projects/graph/src/queue.py
```diff
@@ -0,0 +1,26 @@
...
+            return None
+    
+    def is_Empty(self):
+        if self.size() == 0:
+            return False
+        else:
+            return True
+    
+    def size(self):
+        return len(self.queue)
```
blauente: Good job putting the queue in another class file and importing it. That's the smart way to keep your file sizes down.

projects/graph/src/graph.py
```diff
@@ -1,9 +1,48 @@
...
+        self.vertices = dict() # or {} instead of dict()
+
+    def add_vertex(self, vertex):
+        if vertex not in self.vertices.keys():
+            self.vertices[vertex] = set()
+        return 'vertex already exist...'
+
+    def add_edge(self, start_point, end_point):
+        if start_point in self.vertices and end_point in self.vertices:
+            #graph is bidirectional.. should add edges back and forth
```
blauente: Great work with all the comments in this file, it show that you understand what's going on. 👍

projects/graph/src/graph.py
```diff
@@ -1,9 +1,183 @@
 """
 Simple graph implementation
 """
-
+from queue import Queue
+from stack import Stack
```
blauente: Do you have a `stack` file to import from? I don't have one in your PR. You'll need to create one with a Stack class if not.

blauente: Thank you for pushing the files. The `graph_demo.py` runs beautifully now. 👍

PR #370, opened 2019-01-28
Graphs - Scott Bren

smbren:
Graphs - Scott Bren

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Graphs
+# Graphs -
 
 ## Objectives
 


projects/graph/src/graph.py
```diff
@@ -6,4 +6,13 @@
...
+        self.vertices = {}
+
+    def add_vertex(self, vertex):
+        self.vertices[vertex] = set()
+
+    def add_edge(self, vertex1, vertex2):
+        if vertex1 and vertex2 in self.vertices:
+            self.vertices[vertex1].add(vertex2)
+            self.vertices[vertex2].add(vertex1)
+
```
blauente: Please continue to work on the Stack and Queue classes, and also the BFT, DFT, BFS, DFS, and recursive methods in the `Graph` class.

projects/graph/src/graph.py
```diff
@@ -2,8 +2,96 @@
...
+                for next_node in self.vertices[node]:
+                    s.push(next_node)
+
+    def bfs(self, starting_node, destination_node):
+        
+        queue = [(starting_node, [starting_node])]
+        while queue:
+            (vertex, path) = queue.pop(0)
+            for next in graph[vertex] - set(path):
```
blauente: I think this should be `for next in self.vertices[vertex] - set(path):`. This method should return `True` if the destination node is found, and `False` otherwise.

projects/graph/social/social.py
```diff
@@ -47,6 +47,19 @@ def populateGraph(self, numUsers, avgFriendships):
...
+        sg = SocialGraph()
+        for user in range(numUsers):
+            sg.addUser(f"User {user}")
+        
+        possible_friendships = list(combinations(range(1, len(sg.users)+1), 2))
+        random.shuffle(possible_friendships)
+        friendships = possible_friendships[:15]
+
+        for friendship in friendships:
+            sg.addFriendship(friendship[0], friendship[1])
```
blauente: I don't get anything printing for this file.

projects/graph/src/graph.py
```diff
@@ -2,8 +2,96 @@
...
+
 
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
-        pass  # TODO
+        self.vertices = {}
+
+    def add_vertex(self, vertex):
+        self.vertices[vertex] = set()
```
blauente: The vertices do print when you run `graph_demo.py`. Good work.

PR #371, opened 2019-01-28
Drew Johnson - Graphs

Drewsup123:
@Tulf

### DIFF
--- a/projects/graph/src/graph.py
+++ b/projects/graph/src/graph.py
@@ -1,6 +1,8 @@
 """
 Simple graph implementation
 """
+from structures import Queue
+from structures import LinkedList
 
 
 class Graph:
@@ -17,5 +19,8 @@ class Graph:
     def add_edge(self, vertex1, vertex2):
         if vertex1 and vertex2:
             self.vertices[vertex1].add(vertex2)
+
+    def BGT(self, starting):
+        
 
 
--- a/projects/graph/src/graph.py
+++ b/projects/graph/src/graph.py
@@ -6,4 +6,16 @@ Simple graph implementation
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
-        pass  # TODO
+        self.vertices = {}
+
+    def add_vertex(self, v):
+        if v is not None:
+            self.vertices[v] = set()
+        else:
+            return "Error: Vertex can not be none"
+
+    def add_edge(self, vertex1, vertex2):
+        if vertex1 and vertex2:
+            self.vertices[vertex1].add(vertex2)
+
+
--- a/projects/.gitignore
+++ b/projects/.gitignore
@@ -1,2 +1,3 @@
 .DS_Store
 .vscode/
+.idea/
--- a/projects/graph/.gitignore
+++ b/projects/graph/.gitignore
@@ -2,7 +2,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
-
+.idea/
 # C extensions
 *.so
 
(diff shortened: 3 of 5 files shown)


projects/graph/src/graph.py
```diff
@@ -43,6 +43,31 @@ def dft(self, starting_point, next_set=[], visited=[]):
...
+        # ---------------------------------------------
+        # Create a stack
+        visited = []
+        stack = []
+        # add starting_point to that stack
+        stack.append(starting_point)
+        # while the stack is not None:
+        while len(stack) > 0:
+            # take the first node in stack out
+            visited.append(stack.pop(0))
```
Tulf: :heavy_check_mark:

projects/graph/src/graph.py
```diff
@@ -43,6 +43,31 @@ def dft(self, starting_point, next_set=[], visited=[]):
...
+            # put it in a visited array
+            # if child(ren) is not in the visited array
+            for node in self.vertices[f"{visited[len(visited)-1]}"]:
+                if node not in visited:
+                    stack.append(node)
+                    # print("APPENDED NODE L66 : ", node)
+                    print(node)
```
Tulf: why don't you just print visited? isn't it already in order?

projects/graph/src/graph.py
```diff
@@ -74,8 +75,11 @@ def bfs(self, starting_point, value):
             return True
         return False
 
-    def dfs(self):
-        pass
+    def dfs(self, starting_point, value):
+        a = self.dfts(starting_point)
+        if str(value) in a:
+            return True
+        return False
```
Tulf: I'm not sure this is really a depth first search. And I don't think either of these would work very well for an unidrected graph. But this is certainly interesting not sure I've ever seen a solution like this.

projects/graph/src/graph.py
```diff
@@ -74,8 +75,11 @@ def bfs(self, starting_point, value):
             return True
         return False
 
-    def dfs(self):
-        pass
+    def dfs(self, starting_point, value):
+        a = self.dfts(starting_point)
```
Tulf: this should be able to find a unconnected vertex if I start at it. currently doesn't.

projects/graph/social/social.py
```diff
@@ -46,14 +47,28 @@ def populateGraph(self, numUsers, avgFriendships):
...
+        #     self.addUser(i)
+        # # Create friendships
+        # for j in range(numUsers):
+        #     for k in range(random.randint(0, avgFriendships+1)):
+        #         r_int = random.randint(1, numUsers)
+        #         print("r_int", r_int)
+        #         self.addFriendship(j, r_int)
+
+        # Redone code
+        # adds all users based on the parameter above
```
Tulf: Good it's great that you leave the mistakes in your code. This will be extremely helpful for you when you go back to review since you will be able to access your prior thought process.

projects/graph/social/social.py
```diff
@@ -44,11 +45,15 @@ def populateGraph(self, numUsers, avgFriendships):
...
         # Add users
-
+        for i in range(numUsers):
+            self.addUser(i)
         # Create friendships
+        for j in range(numUsers):
+            for k in range(random.randint(0, avgFriendships+1)):
+                r_int = random.randint(1, numUsers)
+                print("r_int", r_int)
+                self.addFriendship(j, r_int)
```
Tulf: The only suggestion I have here, and this applies overall, is don't commit one liners or spacing changes, that's not very helpful, also try to be a little more specific with your commits. 
I will refer to this guide:
https://github.com/erlang/otp/wiki/writing-good-commit-messages
If you wish to keep a brief title for example you can add comments in the body of the commit.

PR #374, opened 2019-01-28
Eva Li

eeevvaaa:


projects/graph/src/graph.py
```diff
@@ -1,9 +1,86 @@
...
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
-        pass  # TODO
+        self.vertices = {}
+
+    def add_vertex(self, vertex):
+        if vertex not in self.vertices:
+            self.vertices[vertex] = set()
+        else:
+            return
```
blauente: What are you returning?

projects/graph/src/graph_demo.py
```diff
@@ -19,6 +19,12 @@ def main():
     graph.add_edge('0', '3')
     print(graph.vertices)
 
+    print(graph.bft('0'))
+    print(graph.bft('1'))
+    print(graph.dft('0'))
+    print(graph.dft('3'))
+    print(graph.dft_r('0'))
```
blauente: I like that you added these test lines. Good work. 👍

projects/graph/src/graph.py
```diff
@@ -1,9 +1,86 @@
...
+
+    def dft_r(self, starting_vertex, visited = None):
+        if visited == None:
+            visited = []
+        if starting_vertex not in visited:
+            visited.append(starting_vertex)
+        for edge in self.vertices[starting_vertex]:
+            if edge not in visited:
+                self.dft_r(edge, visited)
+        return visited
```
blauente: Looks like everything works here. Very nice!

projects/graph/src/graph.py
```diff
@@ -1,9 +1,86 @@
...
+            current_node = queue.dequeue()
+            visited.append(current_node)
+            for edge in self.vertices[current_node]:
+                if edge not in visited:
+                    queue.enqueue(edge)
+        return visited
+
+    def dft(self, starting_vertex):
+        stack = Stack()
+        visited = []
```
blauente: It's definitely better to use a list here than a set.

PR #375, opened 2019-01-28
Michael Agard

fellanonymous:
@Tulf


projects/graph/src/graph.py
```diff
@@ -20,3 +22,10 @@ def add_edge(self, vertex1, vertex2):
             self.vertices[vertex2].add(vertex1)
         else:
             raise Exception("Edges to nonexistent vetices")
+
+    def bredth_first_traversal(self, start_node):
+        queue = Queue()
```
Tulf: If you're going to use visited to just print out the visited nodes in order i suggest using a list since they have an index.

michaelagard: Based off the code you commented on, do you want me to not use the queue class, instead just use a list? Visited is already a list on line 28.

Tulf: You can use the queue to help guide the bft, I was referring to your visited set.

projects/graph/src/graph.py
```diff
@@ -16,16 +16,23 @@ def add_vertex(self, vertex):
...
             self.vertices[vertex2].add(vertex1)
         else:
             raise Exception("Edges to nonexistent vetices")
 
     def bredth_first_traversal(self, start_node):
+        # Create a new queue using the imported queue file
         queue = Queue()
-        visited = set()
+        # Storage for the nodes that were visted
+        visited = []
```
Tulf: :heavy_check_mark:

Tulf:
This is much better Michael you're committing meaningful changes and really focusing on one issue at a time! :100: :+1:

projects/graph/social/social.py
```diff
@@ -45,10 +47,19 @@ def populateGraph(self, numUsers, avgFriendships):
...
+        for num in range(numUsers):
+            self.addUser(f"User {num}")
         # Create friendships
+        friend_list = []
+        for i in range(numUsers):
+            for j in range(numUsers):
+                if i != j and i < j:
+                    friend_list += [[i, j]]
+        random.shuffle(friend_list)
+        friend_list = friend_list[:10]
```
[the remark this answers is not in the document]
Tulf: Hmm interesting. Could you try to make the relationship 1:1?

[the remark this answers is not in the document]
Tulf: But carry on if it works for now. You can always refactor later.

projects/graph/src/graph.py
```diff
@@ -21,7 +21,7 @@ def add_edge(self, vertex1, vertex2):
         else:
             raise Exception(f'Vertex {vertex1} does not exist.')
 
-    def bredth_first_traversal(self, start_node):
+    def breadth_first_traversal(self, start_node):
```
Tulf: spelling counts :joy:

projects/graph/social/social.py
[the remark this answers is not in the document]
michaelagard: > 
> 
> Hmm interesting. Could you try to make the relationship 1:1?

What do you mean by a 1:1 relationship?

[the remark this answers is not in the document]
Tulf: Why would you have an excess number of friend nodes? the friend list only contains the connections It's the edges of the vertices that is your userID value. That's what I see from how you implemented. So it will grow with respect to how many connections each user has right. You can expect not every user will be friends with everyone, but the friendships should be bidirectional. Because it doesn't seem like to me that you're containing the friend of a friend relationships in friend_list.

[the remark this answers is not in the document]
Tulf: My only concern is that this will cause errors when you try to look at the degrees of seperation.

projects/graph/social/social.py
```diff
@@ -60,7 +62,7 @@ def populateGraph(self, numUsers, avgFriendships):
                 if user_id + 1 != friend_id and user_id < friend_id:
                     friend_list += [[user_id + 1, friend_id]]
         random.shuffle(friend_list)
-        friend_list = friend_list[:10]
+        friend_list = friend_list[:numUsers]
```
Tulf: :100:

projects/graph/social/social.py
```diff
@@ -11,7 +13,7 @@ class SocialGraph:
     def __init__(self):
         self.lastID = 0
         self.users = {}  # Verticies
-        self.friendships = {}  # Edge
+        self.friendships = {}  # Dictionary of users and their edges
```
Tulf: :+1: 
Seems like you get the bidirectional.

projects/graph/social/social.py
```diff
@@ -73,16 +75,26 @@ def getAllSocialPaths(self, userID):
...
+
+            for friend in self.friendships[node]:
+                if friend not in visited:
+
+                    upath = path + [friend]
+
+                    visited[friend] = upath
+
+                    queue.enqueue(upath)
+
```
Tulf: This is excellent work Michael. I'm very happy to see that you went with this implementation strategy for you BFS. I think by adding in these structures you can really start to understand how generalizable these methods really are. All you needed to do here was repeatedly visit the nodes around the passed in node, and then follow the rest of their branches and save the paths. 
This is top notch work and a fantastic turnaround around form last week! I'm really excited to see you make so much improvement.

PR #385, opened 2019-01-28
Simon Elhoej

Elhoej:


projects/graph/src/graph.py
```diff
@@ -1,9 +1,14 @@
...
-        pass  # TODO
+        self.vertices = {}
+
+    def add_vertex(self, value):
+        self.vertices[value] = set()
+
+    def add_edge(self, vert, value):
+        self.vertices[vert].add(value)
+        self.vertices[value].add(vert)
+
```
AI2EJAY: Nice job with getting the initial stages of your graph class set up. 

If you need any help with your traversal methods be sure to reach out. 🙂

PR #389, opened 2019-01-28
Samantha Gatt

samanthagatt:
@Tulf

### DIFF
--- a/projects/graph/src/graph.py
+++ b/projects/graph/src/graph.py
@@ -36,10 +36,20 @@ class Graph:
             if len(q) > 0:
                 self.bft(q.popleft(), q, visited)
 
-    def dft(self, start):
+    def dft_recursive(self, start, visited=set()):
+        if start in self.vertices:
+            print(start)
+            if len(visited) == 0:
+                visited.add(start)
+            for related in self.vertices[start]:
+                if related not in visited:
+                    visited.add(related)
+                    self.dft_recursive(related, visited)
+
+    def dft_stack(self, start):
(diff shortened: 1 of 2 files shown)


projects/graph/src/graph.py
```diff
@@ -6,4 +6,19 @@
...
+        if vertex not in self.vertices:
+            self.vertices[vertex] = set()
+
+    def add_edge(self, vertex_1, vertex_2):
+        if vertex_1 in self.vertices and vertex_2 in self.vertices:
+            self.vertices[vertex_1].add(vertex_2)
+            self.vertices[vertex_2].add(vertex_1)
+        else:
+            if vertex_1 not in self.vertices and vertex_2 not in self.vertices:
+                raise KeyError(f"{vertex_1} and {vertex_2} are not valid vertices")
```
Tulf: Great error checking you specify which vertex is not in the graph, Samantha, and this is all O(1)

projects/graph/src/graph.py
```diff
@@ -24,32 +24,41 @@ def add_edge(self, vertex_1, vertex_2):
...
+            if start in self.vertices:
+                print(start)
+                if not visited:
+                    visited = {start}
+                for related in self.vertices[start]:
+                    if related not in visited:
+                        q.append(related)
+                        visited.add(related)
+                if len(q) > 0:
+                    self.dft_stack(q.pop(), q, visited)
```
Tulf: Someone really loves recursion!

PR #392, opened 2019-01-28
Graphs-I-----JeanetteFernandez

jeanfern5:


jeanfern5:
> Great job implementing all of MVP 🙂
> 
> Having all of those prints can really be useful while visualizing the code and debugging but it can get pretty messy. Python has a handy module that not only solves that problem but offers a lot of other powerful features.
> 
> ### Suggestion
> * I suggest you look into it: https://docs.python.org/2/library/logging.html . Don't worry it's pretty easy to implement and if you prefer videos this one wasn't bad https://www.youtube.com/watch?v=-ARI4Cz-awo

I usually delete my prints so that the final code looks cleaner, but I do sometimes forget. Thank you, I will look into python logging.

PR #393, opened 2019-01-28
Hunter Smith - Graphs

Hunter315:
@Tulf

### DIFF
--- a/projects/graph/src/graph.py
+++ b/projects/graph/src/graph.py
@@ -6,4 +6,27 @@ Simple graph implementation
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
-        pass  # TODO
+        self.vertices = { }
+
+    def add_vertex(self, vertex):
+        self.vertices[vertex] = set()
+
+    def add_edge(self, vertex_a, vertex_b):
+        if vertex_a in self.vertices.keys() and vertex_b in self.vertices.keys():
+            self.vertices[vertex_a].add(vertex_b)
+            self.vertices[vertex_b].add(vertex_a)
+        else:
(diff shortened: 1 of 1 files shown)


projects/graph/src/graph.py
```diff
@@ -6,4 +6,27 @@
...
+
+    def add_vertex(self, vertex):
+        self.vertices[vertex] = set()
+
+    def add_edge(self, vertex_a, vertex_b):
+        if vertex_a in self.vertices.keys() and vertex_b in self.vertices.keys():
+            self.vertices[vertex_a].add(vertex_b)
+            self.vertices[vertex_b].add(vertex_a)
+        else:
+            print("Cannot find all vertices")
```
Tulf: Great work Hunter, you're validating the vertices correctly but try to add some conditional logic to reference a specific vertex in the case where one of them is valid so you can point to the invalid one.

projects/graph/src/graph.py
```diff
@@ -2,6 +2,32 @@
 Simple graph implementation
 """
 
+class Queue:
```
Tulf: Nice try to play around with Python's queue() when you have the time

projects/graph/src/graph_demo.py
```diff
@@ -8,16 +8,34 @@
...
+    graph.add_edge('4', '5')
+    graph.add_edge('3', '4')
+    graph.add_edge('5', '9')
+    graph.add_edge('6', '7')
+    graph.add_edge('7', '9')
+    graph.add_edge('2', '8')
+
     print(graph.vertices)
 
+    graph.bft('2')
```
Tulf: Good idea this definitely makes testing easier.

PR #399, opened 2019-01-29
Brock Rohloff

TheBrockstar:
@Tulf

### DIFF
--- a/projects/graph/src/graph.py
+++ b/projects/graph/src/graph.py
@@ -7,3 +7,4 @@ class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
         pass  # TODO
+ 


projects/graph/src/graph.py
```diff
@@ -2,9 +2,34 @@
...
+            self.vertices[value] = set()
+    
+    def add_edge(self, vertex, target):
+        # print(self.vertices)
+        # print(self.vertices[vertex])
+        # print(hasattr(self.vertices, vertex))
+        if vertex in self.vertices and target in self.vertices:
+            self.vertices[vertex].add(target)
+        else:
+            print('One of the provided vertices does not exist.')
```
Tulf: well or both, but try to add some conditional logic that will reference the specif vertex(ices)

TheBrockstar: Thank you! That has been updated.

Tulf:
What about the remade questions?

PR #403, opened 2019-02-04
Graphs - Evan Carlstrom

ecarlstrom:


ecarlstrom:
Thanks, took a look at add_edge() after your comments earlier and adjusted it based on what you said. I'm testing my traversal methods and this should hopefully fix any issues.

ecarlstrom:
Thanks for the clarification, I'm going to try that manual implementation later as well to make sure I understand the steps. I'll also be trying out the Stack and Queue classes manually rather than using deque for the functions. Just want to make sure I get some practice using both methods.

PR #404, opened 2019-02-04
Andre Myrick - Graphs

AM-Myrick:


codejoncode:
```
for i in range(0, numUsers):
            self.addUser(f"User {i}")
```
Could also be used to generate your names without a random list. 
Hard to understand at first couldn't see that you were in the populateGraph function. definitely, a way to get it done without that function but having a function that serves one purpose is good practice. If there is a problem you can go back to the exact function to fix it.  Final output looks good

codejoncode:
`def populateGraph(self, numUsers, avgFriendships):
        self.lastID = 0
        self.users = {}
        self.friendships = {}
        for i in range(0, numUsers):
            self.addUser(f"User {i}")
        possibleFriendships = []
        for userID in self.users:
            for friendID in range(userID + 1, self.lastID + 1):
                possibleFriendships.append((userID, friendID))
        random.shuffle(possibleFriendships)
        x = 0
        for i in range(0, math.floor(numUsers * avgFriendships / 2)):
            friendship = possibleFriendships[i]
            self.addFriendship(friendship[0], friendship[1])`
This is how we could accomplish everything in the same function. I know you were interested so I thought I share. But again your way works so its not something you have to fix.

PR #452, opened 2019-07-10
Lorenzo-Evans-Graphs

LorenzoEvans:
PR for Graphs.

### DIFF
--- a/projects/graph/graph.py
+++ b/projects/graph/graph.py
@@ -16,7 +16,8 @@ class Graph:
         """
         Add a directed edge to the graph.
         """
-        pass  # TODO
+        self.vertices[v1].add(v2)
+
     def bft(self, starting_vertex):
         """
         Print each vertex in breadth-first order
--- a/projects/graph/graph.py
+++ b/projects/graph/graph.py
@@ -7,11 +7,13 @@ class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
         self.vertices = {}
+        self.queue = Queue()
+
     def add_vertex(self, vertex):
         """
         Add a vertex to the graph.
         """
-        self.vertices[f"{ vertex }"] = set()
+        self.vertices[vertex] = set()
     def add_edge(self, v1, v2):
         """
         Add a directed edge to the graph.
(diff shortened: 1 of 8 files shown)


projects/graph/graph.py
```diff
@@ -7,42 +7,112 @@ class Graph:
...
         Add a vertex to the graph.
         """
-        pass  # TODO
+        self.vertices[vertex] = set()
     def add_edge(self, v1, v2):
         """
         Add a directed edge to the graph.
         """
-        pass  # TODO
+        if v1 not in self.vertices:
```
juliankohlman: Good clean code Lorenzo. If `v1` or our `origin vertex` doesn't exist add it to the graph. Otherwise, add  `v2` our `destination vertex` to `v1` vertices/edges 👍

projects/graph/graph.py
```diff
@@ -7,42 +7,112 @@ class Graph:
...
+            self.add_vertex(v1)
+        else:
+            self.vertices[v1].add(v2)
+
```
juliankohlman: How would you go about refactoring `add_edges()` to work with an `undirected` graph?

projects/graph/graph.py
```diff
@@ -7,42 +7,112 @@ class Graph:
...
     def bft(self, starting_vertex):
         """
         Print each vertex in breadth-first order
         beginning from starting_vertex.
         """
-        pass  # TODO
+
+        visited = []
```
juliankohlman: How would your code change if `visited` was a `Set` instead of a list?

projects/graph/graph.py
```diff
@@ -7,42 +7,112 @@ class Graph:
...
+        vertices = self.vertices
+        queue = self.queue
+        queue.enqueue(starting_vertex)
+        while queue.size():
+            cur_ver = queue.dequeue()
+            if cur_ver not in visited:
```
juliankohlman: A very solid implementation here Lorenzo 👍 All I would do is to add in a print statement before you the `cur_ver` to `visited` so you can check out the traversal in the terminal.

projects/graph/graph.py
```diff
@@ -7,42 +7,112 @@ class Graph:
...
+        self.queue = Queue()
+        return visited
+
     def dft(self, starting_vertex):
         """
         Print each vertex in depth-first order
         beginning from starting_vertex.
         """
-        pass  # TODO
+        stack = self.stack
```
juliankohlman: You may be able to assign something like `stack = Stack()` on line 50. The stack can be independent of the graph instance. Vertices would require the usage of `self` because it's an attribute of the graph class. You can still `push`, `pop` and `iterate` over your Stack without using self, but your code is very explicit and easy to read 👍 Just something to tinker around with. __UPDATE__ I missed your Graph class `init()` code where you added the queue and stack classes. That's a great idea 🔥

PR #455, opened 2019-07-10
Graphs - Michael Trew

CodeDiversity:
@julianKohlman

### DIFF
--- a/projects/graph/graph.py
+++ b/projects/graph/graph.py
@@ -17,43 +17,82 @@ class Graph:
     def add_edge(self, v1, v2):
         if v1 not in self.vertices:
             self.add_vertex(v1)
-        
+
         if v2 not in self.vertices:
             self.add_vertex(v2)
-        
+
         self.vertices[v1].add(v2)
-        
+
         self.vertices[v2].add(v1)
 
     def bft(self, starting_vertex):
-        """
-        Print each vertex in breadth-first order
-        beginning from starting_vertex.
-        """
+        queue = Queue()
+
+        visited = set()
+
+        while queue.size():
+            node = queue.dequeue()
+
+            if not node in visited:
+                visited.add(node)
+
+                for edge in self.vertices[node]:
+                    queue.enqueue(edge)
         pass  # TODO
 
     def dft(self, starting_vertex):
-        """
-        Print each vertex in depth-first order
-        beginning from starting_vertex.
-        """
-        pass  # TODO
+        stack = Stack()
 
-    def dft_recursive(self, starting_vertex):
-        """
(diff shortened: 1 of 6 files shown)


projects/graph/graph.py
```diff
@@ -3,46 +3,58 @@
...
+
     def __init__(self):
         self.vertices = {}
+
     def add_vertex(self, vertex):
-        """
-        Add a vertex to the graph.
-        """
-        pass  # TODO
+        self.vertices[vertex] = set()
```
juliankohlman: Good use of a comment here to explain the _what_ of a `set` Michael 👍 . Depending on the implementation details of your graph you could use a more descriptive name for your `vertex` param. It could be an `id`, `name`, `label` etc..etc...

projects/graph/graph.py
```diff
@@ -3,46 +3,58 @@
...
+        # Sets are unique values, no duplicates
+
     def add_edge(self, v1, v2):
-        """
-        Add a directed edge to the graph.
-        """
-        pass  # TODO
```
juliankohlman: Good job breaking out your logic here Michael. It's _very_ clear what's happening by looking at each condition. How would you go about implementing this if you assumed _both_ `v1` and `v2` _are_ in `self.vertices`? If you were to skip that check you would have to handle the false condition in another way, but you'd also avoid making `add_edge` call on `add_vertex`.

projects/graph/graph.py
```diff
@@ -17,43 +17,82 @@ def add_vertex(self, vertex):
...
         if v2 not in self.vertices:
             self.add_vertex(v2)
-        
+
         self.vertices[v1].add(v2)
-        
+
         self.vertices[v2].add(v1)
 
     def bft(self, starting_vertex):
```
juliankohlman: Your BFT algorithm looks great Michael 👍 , could throw a couple of print statements in there and check out how you example graphs get traversed to solidify your understanding.

projects/graph/graph.py
```diff
@@ -17,43 +17,82 @@ def add_vertex(self, vertex):
...
-        """
-        Print each vertex in depth-first order
-        beginning from starting_vertex.
-        """
-        pass  # TODO
+        stack = Stack()
 
-    def dft_recursive(self, starting_vertex):
-        """
-        Print each vertex in depth-first order
```
juliankohlman: Be sure to print each vertex 👍 , great usage of comments throughout `DFT` and `BFS`, your future self will thank you for this when it comes time to brush up on this for an interview.

juliankohlman (commented):
Code is in good shape, Michael.

PR #456, opened 2019-07-11
Emmanuel Prado

Emmanium:
@KingAtoki

### DIFF
--- a/projects/graph/graph.py
+++ b/projects/graph/graph.py
@@ -3,32 +3,46 @@ Simple graph implementation
 """
 from util import Stack, Queue  # These may come in handy
 
+
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
+
     def __init__(self):
         self.vertices = {}
+
     def add_vertex(self, vertex):
         """
         Add a vertex to the graph.
         """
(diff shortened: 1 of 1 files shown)


projects/graph/graph.py
```diff
@@ -3,56 +3,147 @@
...
+
     def __init__(self):
         self.vertices = {}
+
     def add_vertex(self, vertex):
         """
         Add a vertex to the graph.
         """
-        pass  # TODO
+        self.vertices[vertex] = set()
```
KingAtoki: 👍

projects/graph/graph.py
```diff
@@ -3,56 +3,147 @@
...
+
     def add_edge(self, v1, v2):
         """
         Add a directed edge to the graph.
         """
-        pass  # TODO
+        if v1 not in self.vertices:
```
KingAtoki: You don't want to do this. If the vertex doesn't exist, you should raise an error. This method is strictly to just add an edge. Not a vertex too.

So I would do the opposite of this. If `v1` and `v2` DO exist, then we want to `self.vertices[v1].add(v2)` and `self.vertices[v2].add(v1)`

PR #458, opened 2019-07-11
Jen K - Graphs

pghrogue:
@kingatoki

### DIFF
--- a/projects/graph/graph.py
+++ b/projects/graph/graph.py
@@ -7,35 +7,105 @@ class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
         self.vertices = {}
+
+
     def add_vertex(self, vertex):
         """
         Add a vertex to the graph.
         """
-        pass  # TODO
+        self.vertices[vertex] = set()
+
+
     def add_edge(self, v1, v2):
         """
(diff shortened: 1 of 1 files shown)


projects/graph/graph.py
```diff
@@ -7,35 +7,107 @@ class Graph:
...
-        pass  # TODO
+        self.vertices[vertex] = set()
+
+
     def add_edge(self, v1, v2):
         """
         Add a directed edge to the graph.
         """
-        pass  # TODO
+        if v1 not in self.vertices:
```
KingAtoki: I wouldn't have the add edge method creating verticies if they do not exist. This should be strictly just creating an edge between them. So I would remove the `add` part and raise an error if they do not exist

projects/graph/graph.py
```diff
@@ -7,35 +7,107 @@ class Graph:
...
+            self.add_vertex(v1)
+        if v2 not in self.vertices:
+            self.add_vertex(v2)
+
+        self.vertices[v1].add(v2)
+        #self.vertices[v2].add(v1)
```
KingAtoki: This line is good to be uncommented

pghrogue:
Working out logic with temporary code

PR #459, opened 2019-07-11
Jonathan Brock

jbrock1981:
@KingAtoki

### DIFF
--- a/projects/ancestor/ancestor.py
+++ b/projects/ancestor/ancestor.py
@@ -1,3 +1,111 @@
+class Stack():
+    def __init__(self):
+        self.stack = []
+
+    def push(self, value):
+        self.stack.append(value)
+
+    def pop(self):
+        if self.size() > 0:
+            return self.stack.pop()
+
+        raise IndexError("cannot pop empty stack")
+
+    def size(self):
+        return len(self.stack)
+
+
+class Graph():
+    def __init__(self):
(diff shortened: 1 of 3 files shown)


projects/ancestor/ancestor.py
```diff
@@ -1,3 +1,111 @@
+class Stack():
```
KingAtoki: 👍

projects/ancestor/ancestor.py
```diff
@@ -1,3 +1,111 @@
...
+
+
+class Graph():
+    def __init__(self):
+        self.vertices = {}
+
+    def add_vertex(self, value):
+        self.vertices[value] = set()
+
+    def isvertex(self, value):
```
KingAtoki:
```suggestion
    def is_vertex(self, value):
```
just a suggestion. Looks good though!

projects/ancestor/ancestor.py
```diff
@@ -1,3 +1,111 @@
...
+        return value in self.vertices
+
+    def add_edge(self, v1, v2):
+        if v1 in self.vertices and v2 in self.vertices:
+            self.vertices[v1].add(v2)
+            return
+
+        raise IndexError("verts not found")
```
KingAtoki: I think you could put this in an else block and then you don't need the `return` on line 31. But it's up to you 👍

PR #464, opened 2019-07-11
CSPT2 Jason Hedrick

korghal:
@katiagilligan888

### DIFF
--- a/projects/graph/graph.py
+++ b/projects/graph/graph.py
@@ -11,24 +11,53 @@ class Graph:
         """
         Add a vertex to the graph.
         """
-        pass  # TODO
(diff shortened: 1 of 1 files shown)


katiagilligan888:
Thanks for the initial commit and for creating a PR

katiagilligan888:
Jason, good job so far completing the Day 1 assignment. You should now be moving onto the Day 2 assignment as we will be starting Day 3 tomorrow (Monday 07/15). In the BFT, look at line 40-43. The if statement is wrapping the for loop. You could loop through each edge and if the edge is not visited then enqueue it. This should help resolve the infinite loop.

PR #466, opened 2019-07-11
 Initial commit partial graph completion

Siddartha08:
@KingAtoki

### DIFF
--- a/projects/graph/graph.py
+++ b/projects/graph/graph.py
@@ -3,6 +3,12 @@ Simple graph implementation
 """
 from util import Stack, Queue  # These may come in handy
 
+graph = Graph()
+graph.add_vertex(1)
+# {}
+# {1: set()}
+myset.add(1)
+# set(1)
 class Graph:
     """Represent a graph as a dictionary of vertices mapping labels to edges."""
     def __init__(self):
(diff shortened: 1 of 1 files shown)


projects/graph/graph.py
```diff
@@ -11,37 +17,150 @@ def add_vertex(self, vertex):
         """
         Add a vertex to the graph.
         """
+        self.vertices[vertex] = set()
         pass  # TODO
     def add_edge(self, v1, v2):
         """
         Add a directed edge to the graph.
         """
+        if v1 not in self.vertices:
```
KingAtoki: I wouldn't have the add edge method creating verticies if they do not exist. This should be strictly just creating an edge between them. So I would remove the add part and raise an error if they do not exist

projects/graph/graph.py
```diff
@@ -11,37 +17,150 @@ def add_vertex(self, vertex):
...
+            self.add_vertex(v2)
+
+        if v2 not in self.vertices:
+            self.add_vertex(v1)
+
+        # if you say add an edge from v1 to v2, if you only only do one of the below it will be a directed graph
+        self.vertices[v1].add(v2)
+        # self.vertices[v2].add(v1)
```
KingAtoki: You can uncomment this and it is good 👍

PR #527, opened 2020-02-17
Add test_graph.py file to projects/graph

matt-poloni:
Migrates/translates tests at bottom of `graph.py` into a `test_graph.py` file right next to it in the `projects/graph` directory.


projects/graph/graph.py
```diff
@@ -75,71 +75,3 @@ def dfs_recursive(self, starting_vertex):
         This should be done using recursion.
         """
         pass  # TODO
-
```
br80: keep this

